PR #48, opened 2013-09-20
Change the width of bookmark titles to better use the white space on the screen

dansoap:
you can see the difference here: https://castex.de/stuff/bookmarks.png


dan0xii:
I was searching for outstanding issues regarding the condensed/abbreviated display of the bookmark titles and came across this old discussion. Are there no plans to improve on the layout? I see the work being done on a gridview but I quite like the flat list with title, URL, tag and date. I almost never use the description field and so I have huge amounts of whitespace but can't read many of the bookmarks' titles.

blizzz:
@dan0xii there's some work in (slow) progress https://github.com/nextcloud/bookmarks/pull/402

Issue #113, opened 2014-12-15
Shouldn’t require protocol like http / https


jancborchardt:
Currently when you put in something like duckduckgo.com or owncloud.org – it won’t work. You need to prepend http:// or https://

That should not be required. It needs to be checked if http / https is present, and if not just prepend that automatically.

cc @ttuxx @gaomi @wakeup

TtuxX:
@jancborchardt pull request merged, issue solved

jancborchardt:
@TtuxX awesome, thank you again! :)

TtuxX:
@jancborchardt You are welcome ! Let me know if I can help on similar tasks.

jancborchardt:
@TtuxX cool! If you want to take a stab at fixing the navigation, take a look at https://github.com/owncloud/bookmarks/issues/118 :)

(And are you already in our #owncloud-dev IRC channel?)

PR #114, opened 2014-12-26
Fix http / https protocol requirement.

TtuxX:
Trying to fix issue #113

Fix the http / https prefix requirement when adding a bookmark without prefix by :
1 - Constructing URLs for both protocols,
2 - Retrieving both pages titles,
3 - Choosing the one which works, with a systematic preference for https.

I'd be glad to hear from anyone about several things to discuss :
- I have choosed to always test for the original submitted url, the submitted url with http prefix, and the submitted url with https prefix, but I don't know if that is the best way to achieve tests,
- What about the original (s)ftp bookmark support ? Do you think it is worth keeping it ? in such case, wouldn't it be a bit stupid and heavy to always retrieve each possible page to test each protocol ?
- Should users have their bookmark links always enforced to https if the remote server supports https ? I have currently choosed to respect user's input when they specify a protocol : if you input http://github.com, the bookmark will respect your choice and won't switch to https..
- Problem : In case where https has been implemented on the remote server but the website is not properly configured to work with it, it can end-up obtaining "403 Forbidden" bookmarks titles (experiment it by adding "lefigaro.fr" as bookmark).

Great thanks in advance for sharing your thoughts !! :)

### DIFF
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -70,7 +70,7 @@ class BookmarkController extends ApiController {
 
 		if ($from_own == 0) {
 			// allow only http(s) and (s)ftp
-			$protocols = '/^[hs]{0,1}[tf]{0,1}tp[s]{0,1}\:\/\//i';
+			$protocols = '/^(https?|s?ftp)\:\/\//i';
 			if (preg_match($protocols, $url)) {
 				$datas = Bookmarks::getURLMetadata($url);
 			// if not (allowed) protocol is given, assume http and https (and fetch both)
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -71,8 +71,10 @@ class BookmarkController extends ApiController {
 		if ($from_own == 0) {
 			// allow only http(s) and (s)ftp
 			$protocols = '/^[hs]{0,1}[tf]{0,1}tp[s]{0,1}\:\/\//i';
+			if (preg_match($protocols, $url)) {
+				$datas = Bookmarks::getURLMetadata($url);
 			// if not (allowed) protocol is given, assume http and https (and fetch both)
-			if (! preg_match($protocols, $url)) {
+			} elseif (! preg_match($protocols, $url)) { 
 				// append https to url and fetch it
 				$url_https = 'https://' . $url;
 				$datas_https = Bookmarks::getURLMetadata($url_https);
@@ -80,8 +82,11 @@ class BookmarkController extends ApiController {
 				$url_http = 'http://' . $url;
 				$datas_http = Bookmarks::getURLMetadata($url_http);
 			}
-			// adopt https if it works (switch to http if it doesn't)
-			if (isset($datas_https['title'])) { // test if https works
+			
+			if (isset($datas['title'])) { // prefer original url if working
+				$title = $datas['title'];
+				//url remains unchanged
+			} elseif (isset($datas_https['title'])) { // test if https works
 				$title = $datas_https['title'];
 				$url = $url_https;
 			} elseif (isset($datas_http['title'])) { // otherwise test http for results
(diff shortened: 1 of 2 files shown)


controller/rest/bookmarkcontroller.php
```diff
@@ -68,19 +68,40 @@ public function getBookmarks($type = "bookmark", $tag = '', $page = 0, $sort = "
...
 	public function newBookmark($url = "", $item = array(), $from_own = 0, $title = "", $is_public = false, $description = "") {
 
-		// Check if it is a valid URL
+		if ($from_own == 0) {
+			// allow only http(s) and (s)ftp
+			$protocols = '/^[hs]{0,1}[tf]{0,1}tp[s]{0,1}\:\/\//i';
+			if (preg_match($protocols, $url)) {
+				$datas = Bookmarks::getURLMetadata($url);
+			// if not (allowed) protocol is given, assume http and https (and fetch both)
+			} elseif (! preg_match($protocols, $url)) {
```
TtuxX: @wakeup you're right that's kinda useless here, thanks.

TtuxX:
Thanks  @jancborchardt for your idea
I changed the `addBookmark` function to solve the duplicate issue you have raised. Tell me if it is ok for you :)

Now the `addBookmark` function (from the `controller/lib/bookmarks.php` file) checks wether the url one wants to bookmark exists or not independantly from its protocol.

In order to achieve this, the script excludes the protocol from the DB search. The file `controller/lib/bookmarks.php` now :
- Removes everything from the url before the `://` part (included)
- Uses a `like` statement in the SQL query instead of `=`
- Adds a `%` to the url at the query execution


For the upgrade of old bookmarks from `http` to `https` (when adding the same url), I think that the original idea of @ganomi and you to use dedicated functions seems way cleaner, and that's important in order to avoid messing out with our code...

I am new to github and php, so I let you see for these decisions. If you want to review and change all that, please do so :)

@ganomi Following your message I've started working on your suggestions but didn't ended-up with good code (it turned out for me that the `filter_var` function can not really replace the use of regex when checking if the url protocol is valid).
I'm not the best at php and I am lacking free time so I prefer not switching to complex tasks too fast :)

jancborchardt:
@TtuxX now the protocol is cut off completely – that is problematic since it leads to the links being not secure, as in using http instead of https. If https is put it, it should be saved as https.

jancborchardt:
@TtuxX can you check this out again? :) Thanks!

TtuxX:
@jancborchardt yes sorry for being so long, I am going to try to fix this out soon

jancborchardt:
@TtuxX no worries, and thanks for your contribution!

TtuxX:
Hi @jancborchardt ! I just fixed issues you pointed out.

With these changes, if a user adds an URL specifying its protocol (http/https), bookmarks app will add it respecting his choice.
If a user adds an URL without specifying its protocol, Bookmarks app will sytematically try to add its https version rather than its unsecure one.
If a user adds the same URL twice, adding it with http protocol the first time, and without protocol at all the second time, bookmarks app will upgrade previous bookmark URL from http to https.

Let me know if that is ok for you or/and if you have some suggestions regarding behaviour/code. :)

jancborchardt:
Can you review this @vgezer @MorrisJobke?

controller/rest/bookmarkcontroller.php
```diff
@@ -68,19 +68,40 @@ public function getBookmarks($type = "bookmark", $tag = '', $page = 0, $sort = "
 	 */
 	public function newBookmark($url = "", $item = array(), $from_own = 0, $title = "", $is_public = false, $description = "") {
 
-		// Check if it is a valid URL
+		if ($from_own == 0) {
+			// allow only http(s) and (s)ftp
+			$protocols = '/^[hs]{0,1}[tf]{0,1}tp[s]{0,1}\:\/\//i';
```
MorrisJobke: I would say `/^(https?|s?ftp)\:\/\//i` would be better readable and match more precise

MorrisJobke:
Beside my comment: Nice :+1:

TtuxX:
@MorrisJobke Thanks you ! I changed the regex for the one you proposed. :)

TtuxX:
resolve #113

TtuxX:
Fixes #113

PR #119, opened 2014-12-30
fix add button to show in same line as input field

jancborchardt:
Please review @ganomi @TtuxX @wakeup @owncloud/designers

### DIFF
--- a/css/bookmarks.css
+++ b/css/bookmarks.css
@@ -39,11 +39,7 @@ input.disabled, input.disabled:hover, input.disabled:focus {
 #add_form input {
     margin: 0.3em;
 }
-#add_url {
-    box-sizing: border-box;
-    width: 91.5%;
-    padding-right: 15%;
-}
+
 #bookmark_add_submit {
     position: absolute;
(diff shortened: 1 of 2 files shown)


ganomi:
some changes to make it look neat and tidy

TtuxX:
@jancborchardt Great :+1:

jancborchardt:
@ganomi it was intentional to overlap the field and button to make clear they belong together – can you remove that commit? Also, when the button text is not shown, the loading spinner should show directly in the button.

jancborchardt:
@ganomi hello? Still there? ;) Can you remove your commit with the overlap please as mentioned above?

blizzz:
Does not seem to take effect in current master. I assume that was intended back then for OC 7. The bookmark apps have had a rewrite since then. I try with stable7 branch, but for master and 8.x we need to adjust it. @jancborchardt want to do this?

jancborchardt:
Follow-up at https://github.com/owncloud/bookmarks/pull/170

PR #120, opened 2014-12-31
Adding delicious import and improve tag handling in html import

behrisch:
This adds the possibility to import bookmarks extracted from delicious via the API. The extraction command should look like

 curl https://{USERNAME}:{PASSWORD}@api.del.icio.us/v1/posts/all

Furthermore tags are now parsed from the headlines in a html bookmark file and if a bookmark gets re-added tags will be updated as well.


tflidd:
@behrisch : thanks for all the improvements. One thing which doesn't work very well: If I add a new bookmark via the bookmarklet and I want to add several tags, I can only add one because after adding the first one and hitting enter, the form is automatically submitted.

blizzz:
Hum,  can this be ported to current master? And does it still make sense (i don't know the state of delicious)?

behrisch:
I am not sure about the bookmarklet. I usually separate the tags with comma and it works well then.

behrisch:
I updated the patch now such that it merges cleanly.

blizzz:
I am being told there are still conflicts?

shtrom:
OK, this PR seems to import an XML dump from Delicious (APIv1), but not the HTML file exported from the GUI. It does import the notes properly in this case (helping a bit for #181), but doesn't do it for HTML. Also, it doesn't use the date information, so the  date of all imported bookmarks is set to that of the import.

marcelklehr:
Delicious HTML export files should now import without problems, thus I'm closing this.

PR #133, opened 2015-03-09
tag filter is present again

Chouchen:
override apps.css which have https://github.com/owncloud/core/blob/master/core/css/apps.css#L128

### DIFF
--- a/css/bookmarks.css
+++ b/css/bookmarks.css
@@ -237,7 +237,9 @@ li:hover > .tags_actions > span {
     display:inline;
 }
 li:hover > .tags_actions > em { display : none; }
-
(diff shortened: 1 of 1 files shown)


blizzz:
Works, but I am not convinced by the layout (against current master):

screenshot1

We should just kill the "Filterable Tags" label.

When no tags are assigned, both elements are visible, and that does not make sense either. Only show if at least one tag is existing.

It would also solve #168 but the more consisting design would be as suggest here https://github.com/owncloud/bookmarks/issues/168#issuecomment-119910356 and there is an discussion going on, including how best to filter.

blizzz:
superseded by https://github.com/owncloud/bookmarks/pull/261

PR #151, opened 2015-05-20
Use the title in the input field when adding new bookmark

ege1:
The Original title already got queried and put into the input field. If the user doesn't change it, it will be saved. If the user changes it, he wants it to be changed and not queried again from the original site.

BTW: If a page has two "title" lines, getURLMetadata is too greedy and queries everything from first <title> to the last </title>
You can test that behavior with http://www.metasploitation.com/

This refers to bug https://github.com/owncloud/bookmarks/issues/29

Cheers,

Erik


ege1:
Will create new pull request in new branch

PR #152, opened 2015-05-20
Test title length for new bookmarks

ege1:
Is there a variable where I can find the size limit for that field? I just know that the database has varchar(140) on that field...

This is related to bug: https://github.com/owncloud/bookmarks/issues/29

Cheers,

Erik


ege1:
Will create new pull request in new branch....

PR #153, opened 2015-05-20
Show error message if Title is too long

ege1:
This belongs to pull request: 
https://github.com/owncloud/bookmarks/pull/152
and bug:
https://github.com/owncloud/bookmarks/issues/29

Cheers,

Erik


ege1:
You are right, I'm fixing the wrong definition.

The visibility thing: I do not have any clue, I just copied the definition above about bookmark exists error message... So I think this should work or the other definition wont work either.

I will close this pull requests as I learned it is easier to create a new branch...

I will send a new pull request with all changes.

PR #165, opened 2015-07-08
Add CONTRIBUTING.md

RealRancor:
Shameless copied from https://github.com/owncloud/calendar/pull/839

Ref: https://github.com/owncloud/core/issues/14609

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+## Submitting issues
+
+If you have questions about how to install or use ownCloud, please direct these to the [mailing list][mailinglist] or our [forum][forum]. We are also available on [IRC][irc].
+
+### Short version
+
+ * The [**issue template can be found here**][template]. Please always use the issue template when reporting issues.
+
+### Guidelines
+* Please search the existing issues first, it's likely that your issue was already reported or even fixed.
+  - Go to one of the repositories, click "issues" and type any word in the top search/command bar.
+  - You can also filter by appending e. g. "state:open" to the search string.
+  - More info on [search syntax within github](https://help.github.com/articles/searching-issues)
+* This repository ([bookmarks](https://github.com/owncloud/bookmarks/issues)) is *only* for issues within the ownCloud bookmarks code.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
+## Submitting issues
+
+If you have questions about how to install or use ownCloud, please direct these to the [mailing list][mailinglist] or our [forum][forum]. We are also available on [IRC][irc].
```
blizzz: We could also get an IRC channel of our own, #owncloud-bookmarks as the Mail app has (https://github.com/owncloud/mail/blob/master/CONTRIBUTING.md). What do you think?

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+Help us to maximize the effort we can spend fixing issues and adding new features, by not reporting duplicate issues.
+
+[template]: https://raw.github.com/owncloud/core/master/issue_template.md
+[mailinglist]: https://mailman.owncloud.org/mailman/listinfo/owncloud
+[forum]: https://forum.owncloud.org/
+[irc]: https://webchat.freenode.net/?channels=owncloud&uio=d4
+
+### Contribute Code and translations
+Please check [core's contribution guidelines](https://github.com/owncloud/core/blob/master/CONTRIBUTING.md) for further information about contributing code and translations.
```
blizzz: Maybe add the note that the Contributor Agreement does not need to be signed for when only participating on bookmarks.

blizzz:
My support for shameless actions! :rocket:

CONTRIBUTING.md
RealRancor: Mhhh, the text is a generic text about getting help with the installation or usage of ownCloud so i think #owncloud fits?

blizzz: It's not different in mail. I think, if people look for ownCloud specifically, 
they will end up in core anyway. But to crush into bookmarks is for a reason.

blizzz:
No real blocker → Merging

PR #170, opened 2015-07-10
use proper inline style for bookmark add button and feedback

jancborchardt:
Before (loading feedback showed bottom right below the input field):
capture du 2015-07-10 14-27-48
After (loading feedback shows instead of the + icon):
capture du 2015-07-10 14-27-24

Please review @blizzz @owncloud/designers @ganomi @Chouchen @ege1 @TtuxX

### DIFF
--- a/css/bookmarks.css
+++ b/css/bookmarks.css
@@ -3,9 +3,9 @@ input.disabled, input.disabled:hover, input.disabled:focus {
     background-color: #ddd;
     color: gray;
 }
-#settingsbtn { background: transparent; color:#666; 
+#settingsbtn { background: transparent; color:#666;
                border-bottom: 1px solid #ddd; border-top: 1px solid #fff; }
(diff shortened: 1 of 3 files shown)


blizzz:
Much better! :+1: 

Waiting for other comments. @jonmach your input is also welcome!

TtuxX:
Very nice @jancborchardt ! I like it a lot, it is light and clean !  :)

jonmach:
Sadly, I need to upgrade to 8.1 to test this.  Unfortunately I only found out after testing it, so I seem to have broken something else in the process.  Even my backup failed.  

Ah well.  I will upgrade to 8.1 and pull from the master.  This won't be for a week at least unfortunately

blizzz:
backported to stable8.1 e0becee83861bc378fe6222fdd93379a07838c39

PR #171, opened 2015-07-11
Filter "No TAG"

jforce:
When address lists are huge a filter that allows only see untagged addresses would be very useful.
This feature is particularly relevant in imported lists without tags to allow assign tags to addresses that do not already have.


blizzz:
Um, I don't think you meant to open a pull request, @jforce, did you?

A similar feature request according to your description is https://github.com/owncloud/bookmarks/issues/172

jforce:
i agree
Thanks

PR #179, opened 2015-07-22
Prevent dialog stating "Bookmark added. You can close the window now.…

tmciver:
…" from

displaying when creating a bookmark using bookmarklet. Addresses issue 136.

### DIFF
--- a/js/bookmarklet.js
+++ b/js/bookmarklet.js
@@ -50,10 +50,7 @@ $(document).ready(function () {
 			},
 			success: function (data) {
 				if (data.status === 'success') {
-					$('#bookmarklet_form').html("");
-					OC.dialogs.alert(
-							t("bookmarks", "Bookmark added. You can close the window now."),
-							t("bookmarks", "Bookmark added successfully"), closeWindow, true);
+					window.close();
 				} else {
 					OC.dialogs.alert(t("bookmarks", "Some Error happened."),
 							t("bookmarks", "Error"), null, true);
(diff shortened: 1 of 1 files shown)


blizzz:
@tmciver very good, thank you! One little improvement: we can actually show a success message or green checkmark, so there user really nows it's done. For a second or two (as [recommended here](https://github.com/owncloud/bookmarks/issues/136#issuecomment-119716068)), and then close it automatically.

tmciver:
Sorry for the delay.  Sure, I should be able to add that.  Is there already a way to do this in OC?  I took a quick look at https://github.com/owncloud/core/blob/master/core/js/oc-dialogs.js but it wasn't obvious to me if that sort of dialog was there.

blizzz:
For the record: suggested native  and underscorejs's delay methods on IRC.

blizzz:
> PHP Fatal error:  Uncaught exception 'Symfony\Component\Routing\Exception\RouteNotFoundException' with message 'Unable to generate a URL for the named route "bookmarks.web_view.index" as such route does not exist.' in /home/travis/build/owncloud/core/3rdparty/symfony/routing/Symfony/Component/Routing/Generator/UrlGenerator.php:134

Is what Travis reports. The log is not kept long, though. It does not survive any 5 minutes… @DeepDiver1975 possible to keep it a little bit longer?

Anyway, when I run tests locally, they succeed.

blizzz:
*rebased*

blizzz:
Tests succeed :) @tmciver thank you very much :)

PR #195, opened 2015-09-13
[BUGGED!] Added CRUD actions for REST API

rakshazi:
Hello!
Please, help me with some... bug(?). In public controller I wrote following code:
```php
<?php
public function update($user = null, $password = null, $id = null, $url, $title, $tags = array(), $description, $is_public = false)
        {
                if(!$this->isUser($user, $password))
                {
                        return $this->newJsonErrorMessage("User could not be identified");
                }

                Bookmarks::editBookmark($user,$this->db, $id, $url, $title, $tags, $description, $is_public);

                return new JSONResponse(array('status' => 'ok', 'message' => 'Bookmark has been updated. I hope.'));

        }
```

And added the following in routes.php:
```php
<?php
array('name' => 'public#update', 'url' => '/public/rest/v1/bookmarks/{id}', 'verb' => 'PUT')
```

But when I try to do request (all data correct), I got following message: 
```json
{
  "message": "Current user is not logged in"
}
```
and HTTP status code 401. Explain please, what I doing wrong?

Thank you!

pS: API will be ready after fix of this problem. For now, it works only with bookmarks. No tags available now.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -38,7 +38,10 @@ $application->registerRoutes($this, array('routes' => array(
 		array('name' => 'tags#rename_tag', 'url' => '/tag', 'verb' => 'POST'),
 		array('name' => 'tags#delete_tag', 'url' => '/tag', 'verb' => 'DELETE'),
 		//Public Rest Api
-		array('name' => 'public#return_as_json', 'url' => '/public/rest/v1/bookmark', 'verb' => 'GET'),
+		array('name' => 'public#get', 'url' => '/public/rest/v1/bookmarks', 'verb' => 'GET'),
+		array('name' => 'public#add', 'url' => '/public/rest/v1/bookmarks', 'verb' => 'POST'),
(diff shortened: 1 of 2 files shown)


rakshazi:
@MorrisJobke hi!
Can you help me with this?

MorrisJobke:
> Can you help me with this?

Currently on vacation.

@blizzz Can you have a look?

rakshazi:
Hi guys!
Any update on this?

rakshazi:
@MorrisJobke @blizzz any update on this?

blizzz:
> "Current user is not logged in"

This is thrown, when the @PublicPage annotation is missing (which is the case in update(), which is also correct but wrong with your other methods) and a user was not logged in. If you're interested in the code behind: https://github.com/owncloud/core/blob/master/lib/private/appframework/middleware/security/securitymiddleware.php#L105

My best guess is that your call was done unauthenticated?

rakshazi:
@blizzz yes, this problem was in annotations, thank you for help!

PR #212, opened 2015-11-07
Fix version

DavidPrevot:
Follow up from 2928208f499b191a999e9417969b0b31723a13ce.

Might be too late already, but in case it isn’t…

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -11,5 +11,5 @@
 		<owncloud min-version="8.1" max-version="8.2" />
 	</dependencies>
 	<ocsid>166042</ocsid>
-	<version>0.666666</version>
+	<version>0.6</version>
 </info>


blizzz:
Thanks, but we should not go down. Unfortunately it needs to be 0.670000 :( Otherwise it won't be offered to get it downloaded. What a mess I did :-/

suraia:
My two cents: I somehow ended up with version 0.666666 in my installation (not sure whether I grabbed a snapshot at some point). Therefore, no upgrades ran because all released versions are < 0.666666. Maybe the version should be bumped to 1.x to make sure upgrades run for new versions?

blizzz:
@surala as a quickfix, change the version in your DB. Use command line occ for this:

``sudo -u www-data ./occ config:app:set bookmarks installed_version --value 0.6.0`` from within your Next(/own)cloud directory.  (web user depends on your distribution)

blizzz:
Closing, because the version is already at 0.6.2.

PR #219, opened 2015-12-16
added search bar

ctingwai:
Oh. I didn't see the post, I will go and have a look at the Files app.


blizzz:
There should not be a new search field added, but the one used that is already provided by ownCloud. A small hint https://github.com/owncloud/bookmarks/issues/131#issuecomment-154179849

Issue #187, opened 2015-08-10
Use Mozilla social API for bookmarking


shtrom:
In addition to the bookmarklet, it would be neat if the bookmark application exposed what's needed to be used through the Mozilla social API [0].

[0] https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Social_API

shtrom:
Reusing the bookmarklet does the trick, but some of the sizing needs to be adjusted. Also, I suspect the `output=popup` option is not interpreted by the core anymore, leading to the ownCloud bar being displayed at the top.

![social-api](https://cloud.githubusercontent.com/assets/160280/9172748/f45ea7c6-3fb8-11e5-8a15-33147be39b03.png)

shtrom:
For proper sizing of the panel, we need to be able to add an attribute or style the <body> element of the bookmarklet [0]. I can't work out how to do this with the TemplateResponse construct from the AppFramework, at the moment...

[0] https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Social_API/Share#How_to_properly_size_your_share_panel

shtrom:
A few bugs with social-api-8.2.2:
- The icon is not properly displayed in Firefox
- The window reduces in height whenever a tag is selected

marcelklehr:
Does the social API still exist?

shtrom:
Ha! I doubt it's used much...

I am now using https://github.com/damko/freedommarks-browser-webextension

As far as I am concerned, this ticket can be closed. Feel free to reopen if you think differently.

PR #222, opened 2016-01-07
Social api

shtrom:
For discussion; see also #187.


shtrom:
Wow. That's a dirty branch... Sorry about that.

PR #239, opened 2016-04-17
work around filter_vars limitation with unicode chars in the URL

blizzz:
1. Go to https://de.wikipedia.org/wiki/%C3%84
2. Try to add it to bookmarks via bookmarklet

Currently it will fail silently, a 400 error is returned. It's because filter_var is used for URL verification, alas it returns false if URLs contain non-ascii characters. This works around it, yet i am not sure it's the best solution, only the first that came to my mind. Of course, it's ugly, but better than https://secure.php.net/manual/en/function.filter-var.php#104160.

### DIFF
--- a/controller/lib/helper.php
+++ b/controller/lib/helper.php
@@ -13,4 +13,30 @@ class Helper {
 		}
 	}
 
+	/**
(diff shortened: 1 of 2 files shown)


blizzz:
obsoleted by https://github.com/owncloud/bookmarks/pull/240

PR #261, opened 2016-07-01
Restore broken tag selection functionality

TtuxX:
Hello @blizzz @jancborchardt @Chouchen :)

I am opening this very very small PR in order to fix the current Bookmarks app tag filtering function.

I know and agree the fact that we can (and should) change many many things regarding the OC Bookmarks tag selection behavior and layout (https://github.com/owncloud/bookmarks/issues/118, https://github.com/owncloud/bookmarks/pull/219). 
However, given that the bookmark app development activity is unfortunately extremely low at the moment, one of our top priorities _should be_ to ensure that it _mostly_ works and that there are no main bugs or broken functionalities (e.g.: filter only with 1 tag + need to reload the page to restore the original tag list https://github.com/owncloud/bookmarks/issues/168).

It's quite sad/hard to realise that some basic bookmark app functionalities have been kind of left in a broken state for many months now and that things are not changing. 

This PR contains the minimal amount of code needed to restore the tag selection functionality. I've juste realised that Chouchen wrote an even more basic fix (https://github.com/owncloud/bookmarks/pull/133) 6 months ago but that it has still not been merged.

I understand the need and discussions to improve the tag selection but ... let's fix it first in order to provide every new OC user with a working Bookmarks app version, and then see later on how we will be able to improve it even more !! ? :)

Please let me know your thoughts on this and also if the fix is OK for you ! :)

### DIFF
--- a/js/3rdparty/tag-it.js
+++ b/js/3rdparty/tag-it.js
@@ -323,7 +323,7 @@
 			this._tagInput.val('');
 
 			// insert tag
-			this._tagInput.parent().before(tag);
+			this._tagInput.parent().after(tag);
 		},
 		removeTag: function (tag, animate) {
 			if (typeof animate === 'undefined') {
--- a/css/3rdparty/jquery.tagit.css
+++ b/css/3rdparty/jquery.tagit.css
@@ -3,6 +3,7 @@ ul.tagit {
 	overflow: auto;
     margin-left: inherit; /* usually we don't want the regular ul margins. */
     margin-right: inherit;
+    width: 91.5%;
 }
 ul.tagit li {
 	display: block;
--- a/css/bookmarks.css
+++ b/css/bookmarks.css
@@ -263,6 +263,7 @@ li:hover > .tags_actions > em { display : none; }
 }
 
 #tag_filter a {
+    display: inline-block;
     display: inline;
     float: none;
     margin: 0;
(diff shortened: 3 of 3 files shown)


js/3rdparty/tag-it.js
```diff
@@ -323,7 +323,7 @@
 			this._tagInput.val('');
 
 			// insert tag
-			this._tagInput.parent().before(tag);
+			this._tagInput.parent().after(tag);
```
blizzz: ain't this change go upstream?

TtuxX: What do you mean ? :)

The goal of this modification simply is to change the layout from:

capture
 to 
capture2

blizzz: tag-it.js is a 3rd party lib that we use. We should not change it directly here, but upstream: https://github.com/aehlke/tag-it

However, it looks rather dead, last change 2 yrs ago… Point to you.

blizzz:
Works and improves usage, I agree to merge and improve later on

blizzz:
Thanks for the contribution @TtuxX :)

blizzz:
Backports: https://github.com/owncloud/bookmarks/pull/276 https://github.com/owncloud/bookmarks/pull/277 https://github.com/owncloud/bookmarks/pull/278

TtuxX:
My pleasure @blizzz ! :)

PR #281, opened 2016-08-19
Fix viewBox of SVG to allow proper scaling in Firefox

MorrisJobke:
cc @blizzz 

This should fix the scaling issue in Firefox.


blizzz:
awesome @MorrisJobke thx!

Issue #282, opened 2016-08-19
500 for POST for a new bookmark


theScrabi:
### Steps to reproduce
1. add a new bookmark (does not matter how)

### Expected behaviour
A new bookmark should appere in the list

### Actual behaviour
Nothing happens

### Server configuration
**Operating system**: Ubuntu 16.04

**Web server:** Apache/2.4.18

**Database:** mysql 5.7

**PHP version:** 7.0.8

**ownCloud version:** 9.1

**Bookmarks version:** v0.5.2

**Updated from an older ownCloud or fresh install:**

**Where did you install ownCloud from:**
suse package source

theScrabi:
This is the error i get:
`
[...]

PR #284, opened 2016-08-24
accept pages that return 401-403, show error message on other error c…

blizzz:
…odes

requires forward port to master

solves #282


blizzz:
Forward port in https://github.com/owncloud/bookmarks/pull/285

PR #294, opened 2016-10-11
emit on add/edit/delete

daita:
I am working on [Nextant](https://github.com/daita/nextant), which is a gateway between Nextcloud and [Apache Solr](https://en.wikipedia.org/wiki/Apache_Solr). 

I am thinking about indexing the content from each bookmarks so people can search within their links for anything. I would like to know when a bookmark is created/edited/deleted so I can initiate a background process to index the content of the webpage bookmarked.

This patch add 3 Hooks

### DIFF
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -263,6 +263,8 @@ class Bookmarks {
 			");
 
 		$query->execute(array($id));
+		\OC_Hook::emit('\OCA\Bookmarks', 'post_delete', array('id' => $id));
+		
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


blizzz:
Hey @daita and thanks for your submission!

I am totally happy to include emitting events like this. There is one thing however… the OC_Hooks mechanism is quite old and ugly, because of its static-ness. The nowadays approach is to let listeners register at the Bookmarks app with a Closure, which then will be notified. Would you be open to adjust the code to this approach?

blizzz:
Hey @daita anything about https://github.com/nextcloud/bookmarks/pull/294#issuecomment-253661779?

daita:
Yup, totally forgot about this. I will close and come back with a better pr !

PR #310, opened 2016-12-01
Replace ownCloud with Nextcloud in appinfo

eppfel:
..and remove link to obsolete documentation.

Was removed from Nextcloud Docs in nextcloud/documentation#120

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -2,17 +2,14 @@
 <info>
 	<id>bookmarks</id>
 	<name>Bookmarks</name>
-	<description>Bookmark manager for ownCloud</description>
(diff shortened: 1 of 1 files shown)


blizzz:
It works with both so in the description we could say "Nextcloud and ownCloud" or "Nextcloud and compatible". The other changes are just right!

eppfel:
So, if ownCloud support is planned, we should mention it. If compatibility is a coincidence, I see no need. But not my call.

blizzz:
@eppfel so far the tests are succeeding there as well. I won't put much effort in maintaining that, but as long as it looks OK… in the long run this likely will look different.

blizzz:
👍

PR #315, opened 2016-12-07
adjust authors in info.xml

blizzz:


blizzz:
Failing test was just a time out from a external resource

PR #322, opened 2016-12-16
Add CRUD methods to public REST API

marcelklehr:
Hi,

after my frustrations with the current API (while working on https://github.com/marcelklehr/floccus) I decided to extend the public API with full CRUD support. The methods are mostly copied from the Bookmark controller, as the framework is new to me, so please excuse any bugs :)

### DIFF
--- a/controller/rest/publiccontroller.php
+++ b/controller/rest/publiccontroller.php
@@ -4,6 +4,7 @@ namespace OCA\Bookmarks\Controller\Rest;
 
 use \OCP\AppFramework\ApiController;
 use \OCP\IRequest;
+use \OCP\AppFramework\Http;
 use \OCP\AppFramework\Http\JSONResponse;
 use \OC\User\Manager;
 use OCA\Bookmarks\Controller\Lib\Bookmarks;
@@ -12,6 +13,8 @@ use OCP\Util;
 class PublicController extends ApiController {
 
 	private $userManager;
+	
+	private $userId;
 
 	/** @var Bookmarks */
 	protected $bookmarks;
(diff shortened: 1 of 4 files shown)


marcelklehr:
This is a new take on #218, as that is almost a year old now. No tests, yet. Please provide feedback.

marcelklehr:
(I made the same mistake as the last PR. Now, this PR lets nc handle auth, but still disables CSRF, since it's an API, not a website.)

blizzz:
@marcelklehr cool, thanks for your contribution! I will have a look in detail, nevertheless this is really great! Meanwhile, the tests are failing, could you have a look?

marcelklehr:
Scrutinizer code coverage timed out, some test configurations on travis error with "App "Bookmarks" cannot be installed because it is not compatible with this version of the server." Not sure what that's about.

marcelklehr:
There's still one failing test, I don't get any output or error message though :/

(PS: Sorry for the noise; squashing recommendedTM ;) )

blizzz:
> Scrutinizer code coverage timed out, 

I saw all red yesterday, perhaps this was it then.

> some test configurations on travis error with "App "Bookmarks" cannot be installed because it is not compatible with this version of the server." Not sure what that's about.

Nextcloud master went to 12, which according to our info.xml is not supported, yet. It will also fail, because we are using IDb, which was already deprecated and kicked out. I am working on  replacing that.

Let me deal with making  Nextcloud master work again,  then we can rebase and look further, OK?

marcelklehr:
Ah, ok. Good thing the public controller only uses libbookmarks :)

marcelklehr:
I'm unable to fix the tests atm, it seems. @blizzz you might be quicker in finding the bug. Delete seems to work, so I'm at a loss.. :)

blizzz:
Thanks @marcelklehr so far. I am going to have a look on it, but not sure whether I manage it today.

marcelklehr:
Does testuser actually exist? Looks like it doesn't: https://travis-ci.org/nextcloud/bookmarks/jobs/185447600#L388

marcelklehr:
Phew. Got sucked in way deeper than I wanted to, but it works now. Also I apologize for the spam commits :)

Yamakasi:
Would be great to have this in again it's a missing feature! I hope to see it merged asap :) Thanks!

controller/rest/publiccontroller.php
```diff
@@ -12,15 +13,120 @@
...
+	 * @param string $title
+	 * @param bool $is_public
+	 * @param string $description
+	 * @return JSONResponse
+	 *
+	 * @CORS
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 */
+	public function newBookmark($url = "", $item = array(), $title = "", $is_public = false, $description = "") {
```
blizzz: please update the PHP Doc accordingly ↑

blizzz: Hm, this seems to be the same code as in BookmarkController::newBookmark(), isn't it? We should not duplicate the code, but have it in one place. lib/bookmarks.php could have this perhaps, the instance is already available to both PublicController and BookmarkController.

controller/rest/publiccontroller.php
```diff
@@ -12,15 +13,120 @@
...
+	 * @param null $record_id
+	 * @param string $description
+	 * @return JSONResponse
+	 *
+	 * @CORS
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 */
+	public function editBookmark($record_id = null, $url = "", $item = array(), $title = "", $is_public = false, $description = "") {
+		// Check if it is a valid URL
```
blizzz: same as the comment for  ``newBookmark``

controller/rest/publiccontroller.php
```diff
@@ -12,15 +13,120 @@
...
+	
+	/**
+	 * @param int $id
+	 * @return \OCP\AppFramework\Http\JSONResponse
+	 *
+	 * @CORS
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 */
+	public function deleteBookmark($id = -1) {
```
blizzz: same as the comment for  ``newBookmark``

controller/rest/publiccontroller.php
```diff
@@ -95,4 +201,16 @@ public function newJsonErrorMessage($message) {
 		return new JSONResponse($output);
 	}
 
+	/**
+	 * Checks whether parse_url was able to return proper URL data
+	 *
+	 * @param bool|array $urlData result of parse_url
+	 * @return bool
+	 */
+	protected function isProperURL($urlData) {
```
blizzz: and this is also duplicated

marcelklehr:
@blizzz Thanks for the feedback.
I agree that drier code is good. In this case I wasn't sure how to move that stuff somewhere else. One option would be to simply reference methods of the the bookmarks controller. Moving that stuff into libbookmarks would require differentiation of the different failure cases, (possibly) creating new Exceptions and handling them in the controllers, as libBookmarks probably shouldn't create JSONResponse instances directly. I'd say that's your call :)

blizzz:
@marcelklehr OK, then let's merge it and I will do a subsequent PR with the consolidation. By doing so I will also remove the ridiculous user and password parameters from the GET request (which was there before).

Another thing we could think about is to use the website URL instead of the numeric iD for edit and deletion. Just came to my mind.

blizzz:
Oh and this qualifies to increase versioning to 0.10 I think :) @marcelklehr thanks again for your contribution

marcelklehr:
You're welcome -- it wasn't entirely selfless :) Thanks for merging!

re user and password parameters: I left those in, since they have a usecase: If you don't provide a password you'll get the public bookmarks of user x. Might be worth splitting both use cases into two endpoints, though. I'm not sure.

blizzz:
Yes, you could specify a "public" and a "user" parameter, but splitting this up is better I think. Anyhow, there is no built-in way to create public bookmarks. Well, until this PR :)

marcelklehr:
Come to think about it, bookmarkcontroller could probably be fully replaced by the public controller, can it not?

PR #328, opened 2017-01-09
enabled preflighted CORS method 

damko:
I wasn't able to connect my [chrome extension](https://github.com/damko/nextcloud_bookmarks_chrome) to the API because of some CORS problems
By reading the [manual](https://docs.nextcloud.com/server/11/developer_manual/app/api.html) I found out that this line is required for CORS to work properly.

### DIFF
--- a/controller/rest/publiccontroller.php
+++ b/controller/rest/publiccontroller.php
@@ -20,7 +20,8 @@ class PublicController extends ApiController {
 	protected $bookmarks;
 
 	public function __construct($appName, IRequest $request, $userId, Bookmarks $bookmarks, Manager $userManager) {
-		parent::__construct($appName, $request);
+		//see https://docs.nextcloud.com/server/11/developer_manual/app/api.html [Modifying the CORS header]
+		parent::__construct($appName, $request, 'PUT, POST, GET, DELETE, OPTIONS');
 
 		$this->bookmarks = $bookmarks;
 		$this->userManager = $userManager;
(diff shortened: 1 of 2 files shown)


controller/rest/publiccontroller.php
```diff
@@ -40,7 +41,7 @@ public function __construct($appName, IRequest $request, $userId, Bookmarks $boo
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 */
-	public function newBookmark($url = "", $item = array(), $title = "", $is_public = false, $description = "") {
```
blizzz: I think this is a breaking change, also that one in line 76

controller/rest/publiccontroller.php
```diff
@@ -93,7 +94,7 @@ public function newBookmark($url = "", $item = array(), $title = "", $is_public
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 */
-	public function editBookmark($record_id = null, $url = "", $item = array(), $title = "", $is_public = false, $description = "") {
+	public function editBookmark($record_id = null, $url = "", $tags = array(), $title = "", $is_public = false, $description = "") {
```
blizzz: here a breaking change as wellthis and the change below makes the API

blizzz:
Thank you for your contribution and sorry for the late reply! I think the parameter changes causes the tests to fail, also are not needed for the preflighted CORS support. Or am I wrong?

marcelklehr:
Fixed by #338

PR #332, opened 2017-01-30
Use query builder

### LINKED ISSUE #305: TAG rename doesn't work
### Steps to reproduce
1. Create a bookmark
2. Add a tag to the bookmark
3. Edit the tag changing its name

### Expected behaviour
The tag get renamed

### Actual behaviour
Nothing happens.
An error is printed in nextcloud.log

### Server configuration
**Operating system**: debian 8

**Web server:** apache2 v2.4

**Database:** mariadb 10.0.28

**PHP version:** 5.6.27

**nextcloud version:** 10.0.1

**Bookmarks version:** 0.9.0

**Updated from an older nextcloud or fresh install:** fresh

**Where did you install ownCloud from:**

I attach nextcloud.log with the specific error.
[nextcloud.log.zip](https://github.com/nextcloud/bookmarks/files/594023/nextcloud.log.zip)

marcelklehr:
This is a work in progress, I've added you as a collaborator so you can also push to the branch @blizzz ;)

This will hopefully also fix #305

### DIFF
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -76,19 +76,19 @@ class Bookmarks {
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->select('t.tag')
-		->selectAlias($qb->createFunction('COUNT(bookmark_id)'), 'nbr')
-		->from('bookmarks_tags', 't')
-		->innerJoin('t','bookmarks','b',$qb->expr()->andX(
-			$qb->expr()->eq('b.id', 't.bookmark_id'),
-			$qb->expr()->eq('b.user_id', $qb->createNamedParameter($userId));
+			->select('t.tag')
+			->selectAlias($qb->createFunction('COUNT(bookmark_id)'), 'nbr')
+			->from('bookmarks_tags', 't')
+			->innerJoin('t','bookmarks','b',$qb->expr()->andX(
+				$qb->expr()->eq('b.id', 't.bookmark_id'),
+				$qb->expr()->eq('b.user_id', $qb->createNamedParameter($userId))));
 		if (!empty($filterTags)) {
 			$qb->where($qb->expr()->notIn('t.tag', $filterTags));
 		}
 		$qb
-		->groupBy('t.tag')
-		->orderBy('nbr', 'DESC')
-		->setFirstResult($offset);
+			->groupBy('t.tag')
+			->orderBy('nbr', 'DESC')
+			->setFirstResult($offset);
 		if ($limit != -1) {
 			$qb->setMaxResults($limit);
 		}
@@ -106,18 +106,18 @@ class Bookmarks {
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->select('*')
-		->from('bookmarks')
-		->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId))
-		->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($id));
+			->select('*')
+			->from('bookmarks')
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($id)));
 		$result = $qb->execute()->fetch();
 		
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->select('tag')
-		->from('bookmarks_tags')
-		->where($qb->expr()->eq('bookmark_id', $qb->createNamedParameter($id));
+			->select('tag')
+			->from('bookmarks_tags')
+			->where($qb->expr()->eq('bookmark_id', $qb->createNamedParameter($id)));
 		$result['tags'] = $qb->execute()->fetchColumn();
 		return $result;
 	}
@@ -133,10 +133,10 @@ class Bookmarks {
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->select('id')
-		->from('bookmarks')
-		->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId))
-		->andWhere($qb->expr()->eq('url', $qb->createNamedParameter($encodedUrl));
+			->select('id')
+			->from('bookmarks')
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('url', $qb->createNamedParameter($encodedUrl)));
 		$result = $qb->execute()->fetch();
 		if ($result) {
 			return $result['id'];
@@ -201,10 +201,10 @@ class Bookmarks {
 		}
 
 		$qb
-		->from('bookmarks', 'b')
-		->innerJoin('b', 'bookmarks_tags', 't', 't.bookmark_id = b.id')
-		->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userid))
-		->groupBy('b.id');
+			->from('bookmarks', 'b')
+			->innerJoin('b', 'bookmarks_tags', 't', 't.bookmark_id = b.id')
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userid)))
+			->groupBy('b.id');
 
 		if ($public) {
 			$qb->andWhere('public = 1');
@@ -256,7 +256,7 @@ class Bookmarks {
 			$filterExpressions[] = $qb->expr()->eq('t.tag', $qb->createNamedParameter($filter));
 			if (!$filterTagOnly) {
 				foreach ($otherColumns as $col) {
-					$filterExpressions[] = $qb->expr()->like($qb->createFunction('lower(' . $col . ')', $qb->createNamedParameter('%' . strtolower($filter) . '%'));
+					$filterExpressions[] = $qb->expr()->like($qb->createFunction('lower(' . $col . ')', $qb->createNamedParameter('%' . strtolower($filter) . '%')));
 				}
 			}
 		}
@@ -280,10 +280,10 @@ class Bookmarks {
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->select('id')
-		->from('bookmarks')
-		->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId))
-		->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($id))
+			->select('id')
+			->from('bookmarks')
+			->where($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($id)));
 
 		$id = $qb->execute()->fetchColumn();
 		if ($id === false) {
@@ -293,15 +293,15 @@ class Bookmarks {
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->delete('bookmarks')
-		->where($qb->expr()->eq('id', $qb->createNamedParameter($id))
+			->delete('bookmarks')
+			->where($qb->expr()->eq('id', $qb->createNamedParameter($id)));
 		$qb->execute();
 
 		$qb = $this->db->getQueryBuilder();
 		$qb->automaticTablePrefix(true);
 		$qb
-		->delete('bookmarks_tags')
-		->where($qb->expr()->eq('bookmark_id', $qb->createNamedParameter($id))
+			->delete('bookmarks_tags')
+			->where($qb->expr()->eq('bookmark_id', $qb->createNamedParameter($id)));
 		$qb->execute();
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


controller/lib/bookmarks.php
```diff
@@ -72,25 +72,23 @@ public function __construct(
...
-		$tags = $query->fetchAll();
+    		$qb->where($qb->expr()->notIn('tag', $filterTags));
+    	}
+		$qb
+		->groupBy('tag')
+		->orderBy('nbr', 'DESC')
+		->setFirstResult($offset)
+		->setMaxResults($limit);
+		$qb->setParameter(':user_id', $userId);
+    	$tags = $qb->execute()->fetchAll();
```
marcelklehr: Not sure this will work. Pls, correct, if not :)

controller/lib/bookmarks.php
```diff
@@ -72,25 +72,23 @@ public function __construct(
...
 	 */
 	public function findTags($userId, $filterTags = [], $offset = 0, $limit = -1) {
-		$params = array_merge($filterTags, $filterTags);
-		array_unshift($params, $userId);
-		$notIn = '';
+		$qb = $this->db->getQueryBuilder();
+		$qb->automaticTablePrefix(true);
+		$qb
+		->select('tag', $qb->createFunction('COUNT(*)'))
+		->selectAlias('COUNT(*)', 'nbr')
```
marcelklehr: This doesn't do what it's supposed to:
```
    0 => Array (
         'tag' => 'cloud'
-        'nbr' => 1
+        'nbr' => 'COUNT(*)'
+        'COUNT(*)' => '1'
```

controller/lib/bookmarks.php
```diff
@@ -397,35 +416,45 @@ public function editBookmark($userid, $id, $url, $title, $tags = [], $descriptio
...
-		$result = $query->execute($params);
 
+		$qb = $this->db->getQueryBuilder();
+		$qb->automaticTablePrefix(true);
+		$qb
+		->update('bookmarks', 'bm')
+		->set('bm.url', ':url')
+		->set('bm.title', ':title')
+		->set('bm.public', ':public')
+		->set('bm.description', ':description')
```
marcelklehr: How is one supposed to supply values here? https://github.com/nextcloud/server/blob/master/lib/private/DB/QueryBuilder/QueryBuilder.php#L660 says parameters are allowed, but [apparently this doesn't work](https://travis-ci.org/nextcloud/bookmarks/jobs/197972122#L395).
Supplying values directly also doesn't work, because they're quoted like column names (see the url value): https://travis-ci.org/nextcloud/bookmarks/jobs/197803504#L395

marcelklehr: opened https://github.com/nextcloud/server/issues/3373 for this

controller/lib/bookmarks.php
```diff
@@ -315,47 +328,37 @@ public function deleteUrl($userId, $id) {
...
-
-			$params = [$new, $old, $userId];
-			$query->execute($params);
-		}
+		// Remove potentially duplicated tags
+		$qb = $this->db->getQueryBuilder();
+		$qb->automaticTablePrefix(true);
+		$qb
+		->delete('bookmarks_tags', 'tgs')
+		->innerJoin('bm', 'bookmarks', 'tgs.bookmark_id = bm.id')
```
marcelklehr: Mmh, a quick search spit out these, though: [[1]](https://stackoverflow.com/questions/2114534/mysql-syntax-for-join-update/2114583#2114583) [[2]](https://stackoverflow.com/questions/652770/delete-with-join-in-mysql#4192849)

blizzz:
> This is a work in progress, I've added you as a collaborator so you can also push to the branch @blizzz ;)

Ty! As contributor, you should also be able to create and push to branches within the bookmark repo.

blizzz:
I have some fixes, will add more and push later, but first kiddo wants to jump out of the bathtub ;)

blizzz:
Pushed, essentially it is how you passed the parameters. There are also some style issues, e.g. indentation, and we use the short way for arrays, etc.

I did not do a complete review yet, but the tests should pass now.

blizzz:
> but the tests should pass now.

so much about the theory… locally they pass…

blizzz:
@marcelklehr in a nutshell, every comparison should go through the expression builder. E.g.

```php
->where('tgs.tag = :tag')
```

would be

```php
->where($qb->expr()->eq('tag', $qb->createNamedParameter($old))
```

Or  ``createParameter('name')``  instead of ``createNamedParameter``, which i used in my commit.

This causes most problems. Would you have the time to change this accordingly?

marcelklehr:
Thanks! Ah, ok, I will change that

marcelklehr:
All tests pass, now, except for owncloud master! :tada:

controller/lib/bookmarks.php
```diff
@@ -315,47 +316,28 @@ public function deleteUrl($userId, $id) {
...
+			->andWhere($qb->expr()->eq('t.tag', $qb->createNamedParameter($old)));
+		$qb->execute();
+
+		// Update tags to the new label
+		$qb = $this->db->getQueryBuilder();
+		$qb->automaticTablePrefix(true);
+		$qb
+			->update('bookmarks_tags', 'tgs')
+			->set('tgs.tag', $qb->createNamedParameter($new))
+			->innerJoin('bm', 'bookmarks', 'tgs.bookmark_id = bm.id')
```
marcelklehr: Not allowed by mysql or not allowed by the query builder?

controller/lib/bookmarks.php
```diff
@@ -397,35 +376,34 @@ public function editBookmark($userid, $id, $url, $title, $tags = [], $descriptio
...
 
+		$qb = $this->db->getQueryBuilder();
+		$qb->automaticTablePrefix(true);
+		$qb
+			->update('bookmarks')
+			->set('url', $qb->createNamedParameter(htmlspecialchars_decode($url)))
+			->set('title', $qb->createNamedParameter(htmlspecialchars_decode($title)))
+			->set('public', $qb->createNamedParameter($isPublic))
+			->set('description', $qb->createNamedParameter(htmlspecialchars_decode($description)))
+			->set('lastmodified', $qb->createFunction('UNIX_TIMESTAMP()'))
```
marcelklehr: Tests seem to pass for all. For SQLite this is [automatically translated](https://travis-ci.org/nextcloud/bookmarks/jobs/200094785#L394) to "strftime('%s','now')"

marcelklehr:
@blizzz re pushing to branches: In fact, being a contributor in github's terms just means that I authored commits that ended up in a branch in the main repo, not necessarily that I have permission to push new ones myself ;)

blizzz:
great work @marcelklehr i try to give it a final look tonight

controller/lib/bookmarks.php
```diff
@@ -457,46 +421,70 @@ public function addBookmark($userid, $url, $title, $tags = array(), $description
...
-			$query->execute($params);
+
+			$qb
+				->where($qb->expr()->like('url', $qb->createParameter('compareUrl'))) // Find url in the db independantly from its protocol
+				->andWhere($qb->expr()->eq('user_id', $qb->createParameter('userID')));
+				$qb->setParameters([
+					'userID' => $userid,
+					'url' => $decodedUrl,
+					'compareUrl' => '%' . $this->db->escapeLikeParameter($decodedUrlNoPrefix),
+					'title' => $title,
```
marcelklehr: Should $title be htmlspecialchars_decode`d here, too, like on L480?

controller/lib/bookmarks.php
```diff
@@ -230,44 +235,35 @@ public function findBookmarks(
...
-					$sql .= ' AND lower( concat(url,title,description,IFNULL(tags,\'\') )) like ? ';
-				} else {
-					$sql .= ' AND lower(url || title || description || IFNULL(tags,\'\') ) like ? ';
+		$filterExpressions = [];
+		$otherColumns = ['b.url', 'b.title', 'b.description'];
+		foreach ($filters as $filter) {
+			$filterExpressions[] = $qb->expr()->eq('t.tag', $qb->createNamedParameter($filter));
+			if (!$filterTagOnly) {
+				foreach ($otherColumns as $col) {
+					$filterExpressions[] = $qb->expr()->like($qb->createFunction('lower(' . $qb->getColumnName($col) . ')',
```
blizzz: ``like`` misses parameters

blizzz: tests succeed, but I find it suspicious, perhaps a misplaced bracket? Could you check?

controller/lib/bookmarks.php
```diff
@@ -166,58 +170,59 @@ public function findBookmarks(
...
+			$qb->setMaxResults($limit);
+			if ($offset != null) {
+				$qb->setFirstResult($offset);
+			}
 		}
 
-		$query = $this->db->prepare($sql, $limit, $offset);
-		$query->execute($params);
-		$results = $query->fetchAll();
+		$results = $qb->execute()->fetchAll();
```
blizzz: the query does not return bookmarks without tags

controller/lib/bookmarks.php
blizzz: Seems to work either way for me. Leave it out then 🤷

controller/lib/bookmarks.php
```diff
@@ -230,44 +235,35 @@ public function findBookmarks(
...
 				}
-				$params[] = '%' . strtolower($filter) . '%';
 			}
 		}
+		if ($connectWord == 'AND') {
+			$filterExpression = call_user_func_array([$qb->expr(), 'andX'], $filterExpressions);
+		}else {
+			$filterExpression = call_user_func_array([$qb->expr(), 'orX'], $filterExpressions);
+		}
+		$qb->andWhere($filterExpression);
```
marcelklehr: ahh, this is the offending line. It should only be run, if there actually are any filters :)

marcelklehr:
I couldn't find anything wrong with the editBookmark or addTags methods, so I suspect findBookmarks' aggregate function is behaving wrongly and every bookmark is getting all tags. Can you confirm that?

controller/lib/bookmarks.php
marcelklehr: the brackets are a bit confusing here :) the statement continues on the next line. That's where the second parameter is ;)

blizzz: Sure, however ``createFunction`` is not closed here.

controller/lib/bookmarks.php
marcelklehr: Fixed above: https://github.com/nextcloud/bookmarks/pull/332/files#diff-60bf617da66ab938849bb119ba377bc5R249

marcelklehr:
Once you say it's good to go, I'll squash these (unless you're doing a squash merge anyway?)

marcelklehr:
Found the bug in editBookmark and fixed it. I also expanded the associated test case to catch this kind of error in the future.

blizzz:
https://github.com/nextcloud/bookmarks/pull/332#discussion_r101903132

This is still an issue. To reproduce:
1. Add a bookmark (do not assign any tags)
2. → it will appear on the list
3. Reload → It is missing

marcelklehr:
@blizzz ah, sorry I misinterpreted that comment and inadvertently found a different bug :'D The problem was an inner join instead of a left join. Fixed and expanded test case.

The new renameTag test fails now.

marcelklehr:
Fixed renameTag.

blizzz:
It's getting better and better, I am really excited :)

One more thing however

1. Have two bookmarks A and B with tags "tag1 tag2" for A and "tag1" for B.
2. Filter for "tag1"
→ Only "tag1" is shown for Bookmark A, but all are expected.

marcelklehr:
I'm excited, too :D 
That tags bug should be fixed now.

marcelklehr:
@nickvergessen Are you a maintainer as well? In that case could you take a look at it again?

blizzz:
@marcelklehr [this](https://github.com/nextcloud/bookmarks/pull/332#issuecomment-284267090) is fixed, however when when I filter for two tags, the results do not shop up.  For example:

Have a bookmark with tags "tag1" "tag2". Filter for "tag1" → it is shown. Filter additionally for "tag2" → bookmark is not shown.

Looks like the resulting query wants the tag to have filtered values at the same time:

spectacle zm8208

So I guess the main issue is that there is just one tag per row that can be compared. 

What works for me is to OR-connect the filters and filter again on the resulting tags by regexp… ``HAVING  (tags REGEXP 'tag1$' OR tags REGEXP 'tag1,') AND (tags REGEXP 'tag2$' OR tags REGEXP 'tag2,')`` but it feels like a very bad and ugly hack to make it work with the current from/join layout. Perhaps you've a better idea?

marcelklehr:
Ah, gotcha. We could do a join for every tag to filter by, that's basically what the old implementation did (only with an EXISTS sub query). A join a day keeps the doctor away...

blizzz:
Now phpunit started running amok for reasons… i created a seperate PR here https://github.com/nextcloud/bookmarks/pull/342 *fingers crossed*

blizzz:
@marcelklehr could you rebase? the tests should succeed then.

blizzz:
@marcelklehr awesome, thx! 🚀 

A sidenote: Github does not inform/notify about rebases, otherwise i would have seen and merged this earlier.

marcelklehr:
:tada: yay!

ah, ok. sorry :)

PR #335, opened 2017-02-09
add screenshots to info.xml

adsworth:
I guess the syntax is OK, but I couldn't test as the appstore doesn't allow just info.xml validation.


Henni:
executing
```
wget https://apps.nextcloud.com/schema/apps/info.xsd
xmllint appinfo/info.xml --schema info.xsd --noout
```
leads to:
```
info.xml:8: element licence: Schemas validity error : Element 'licence': This element is not expected. Expected is one of ( description, version ).
info.xml fails to validate
```

adsworth:
@Henni that's weird as the bookmarks app is already on the app store.

adsworth:
info.xml is now validated against the appstores info.xml. For some reason it didn't like the productivity category, so I changed that to organization.

blizzz:
👍 thanks guys

PR #337, opened 2017-02-10
Test renameTag

marcelklehr:
As per  #305 this test is expected to fail on mysql and pgsql atm. ~~Please hold off merging until #332 is merged, which should fix this.~~ This should probably be merged before #332, so we can verify just that ;)

### DIFF
--- a/tests/lib_bookmark_test.php
+++ b/tests/lib_bookmark_test.php
@@ -84,6 +84,45 @@ class Test_LibBookmarks_Bookmarks extends TestCase {
 		$this->libBookmarks->addBookmark($this->userid, 'http://nextcloud.com', 'Nextcloud project', array('oc', 'cloud'), 'An awesome project');
 		$this->assertEquals(array(0 => array('tag' => 'cloud', 'nbr' => 1), 1 => array('tag' => 'oc', 'nbr' => 1)), $this->libBookmarks->findTags($this->userid));
 	}
+  
+	function testRenameTag() {
+		$this->cleanDB();
+		$secondUser = $this->userid . "andHisClone435";
+		$this->libBookmarks->addBookmark($this->userid, "http://www.google.de", "Google", array("one"), "PrivateNoTag", false);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.9gag.com", "9gag", array("two", "three"), "PublicTag", true);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.google.de", "Google", array("one"), "PrivateNoTag", false);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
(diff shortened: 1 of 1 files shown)


marcelklehr:
Note that the test fails on postgres as well atm, because it sorts tags with the same number of bookmarks differently :/

blizzz:
> Note that the test fails on postgres as well atm, because it sorts tags with the same number of bookmarks differently :/

Perhaps we could assert by the presence instead of order (or enforce an order)?

tests/lib_bookmark_test.php
```diff
@@ -84,6 +84,46 @@ function testFindTags() {
...
+		$this->libBookmarks->addBookmark($this->userid, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.9gag.com", "9gag", array("two", "three"), "PublicTag", true);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.google.de", "Google", array("one"), "PrivateNoTag", false);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.9gag.com", "9gag", array("two", "three"), "PublicTag", true);
+		
+		$firstUserTags = $this->libBookmarks->findTags($this->userid);
+		$this->assert(in_array(array('tag' => 'one', 'nbr' => 2), $firstUserTags);
```
blizzz: one bracket is missing causing a syntax error.

It's also a good idea to use the short ``[…]`` to declare an array instead of of` ugly ``array(…)``.

marcelklehr: aah, right. I always forget about that :)

tests/lib_bookmark_test.php
```diff
@@ -84,6 +84,46 @@ function testFindTags() {
...
+		$this->libBookmarks->addBookmark($this->userid, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.9gag.com", "9gag", array("two", "three"), "PublicTag", true);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.google.de", "Google", array("one"), "PrivateNoTag", false);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
+		$this->libBookmarks->addBookmark($secondUser, "http://www.9gag.com", "9gag", array("two", "three"), "PublicTag", true);
+		
+		$firstUserTags = $this->libBookmarks->findTags($this->userid);
+		$this->assert(in_array(['tag' => 'one', 'nbr' => 2], $firstUserTags));
```
blizzz: Oh, i oversaw this one at first time. ``assert`` itself does not exist, you need ``assertTrue()`` here and in the following lines.

marcelklehr:
As per https://github.com/nextcloud/bookmarks/issues/305#issuecomment-269533294 the current failure on pgsql is expected as well -- forgot about that.

blizzz:
👍 thx @marcelklehr :)

PR #338, opened 2017-02-10
Consolidate private with preliminary public api

### LINKED ISSUE #330: Support app/device password in REST API
The REST API for getting the bookmarks seems to check the password in way so it doesn't support a generated app/device password. Can this be changed easily? Or can I help implementing this?

### Steps to reproduce
1. Generate new app/device password on your account page
2. call https://example.com/apps/bookmarks/public/rest/v1/bookmark?user=[user]&password=[password] with the above generated password

### Expected behaviour
List of bookmarks is returned.

### Actual behaviour
Error "Wrong password for user [user]" is returned.

**Nextcloud version:** 11.0.0

**Bookmarks version:** 0.9.1

marcelklehr:
This basically changes access to the private bookmarks and tags API to everyone without CSRF tokens and with CORS support, removes the public controller, and adjusts the public controller tests to work with the bookmark controller.

This is a work-in-progress and a follow-up on #322

Fixes #330, #328, #2


marcelklehr:
I merged public_controller#return_as_json() into bookmarkcontroller#getBookmarks. As a result there's some duplicate functionality / redundant parameters now. This is to keep compatibility with the client-side code. If you want, I can adjust the client-side code to use the new parameters and remove the old ones.

Nonetheless, all tests are passing now :)

appinfo/routes.php
```diff
@@ -35,11 +35,6 @@
 		array('name' => 'tags#full_tags', 'url' => '/tag', 'verb' => 'GET'),
 		array('name' => 'tags#rename_tag', 'url' => '/tag', 'verb' => 'POST'),
 		array('name' => 'tags#delete_tag', 'url' => '/tag', 'verb' => 'DELETE'),
-		//Public Rest Api
-		array('name' => 'public#return_as_json', 'url' => '/public/rest/v1/bookmark', 'verb' => 'GET'),
-		array('name' => 'public#new_bookmark', 'url' => '/public/rest/v1/bookmark', 'verb' => 'POST'),
-		array('name' => 'public#edit_bookmark', 'url' => '/public/rest/v1/bookmark/{id}', 'verb' => 'PUT'),
-		array('name' => 'public#delete_bookmark', 'url' => '/public/rest/v1/bookmark/{id}', 'verb' => 'DELETE'),
```
blizzz: Totally removing them would kill compatibility. Better if we could deprecate them over a couple releases.

controller/rest/bookmarkcontroller.php
```diff
@@ -59,25 +62,77 @@ public function legacyGetBookmarks($type = "bookmark", $tag = '', $page = 0, $so
...
+		$page = 0,
+		$sort = "bookmarks_sorting_recent", // legacy
+		$user = null,
+		$tags = array(),
+		$conjunction = "or",
+		$select = null,
+		$sortby = "",
+		$search = array()
+	) {
+		if ($user == null) {
```
blizzz: always use === as equal comparison parameter

appinfo/routes.php
marcelklehr: we don't need to keep the newly added CRUD methods, though, right? They haven't made it into any release yet, afaik

blizzz: Yes, what wasn't there in 0.9.1 is vogelfrei.

appinfo/routes.php
```diff
@@ -35,12 +35,9 @@
...
 		//Legacy Routes
+		array('name' => 'public#return_as_json', 'url' => '/public/rest/v1/bookmark', 'verb' => 'GET'),
+
```
marcelklehr: Is this ok?

blizzz:
I rebased this locally. It seems something is wrong (perhaps with the annotated checks), because I am being logged out when access the bookmarks page normally via web interface.

blizzz:
there lies the issue i think

```
{
  …"message":"Undefined index: PHP_AUTH_USER at \/srv\/http\/nextcloud\/master\/lib\/private\/AppFramework\/Middleware\/Security\/CORSMiddleware.php#88"
…
```
not the index, but that we are logged out and have no provided parameters for logging in.

marcelklehr:
So, the problem I presume is that the client doesn't send auth data? Shouldn't there be a session cookie being sent along?

blizzz:
The CORS annotation makes sure that sessions are not being used, therefore the logout happens.

marcelklehr:
So, how do I realize an API that should be able to be used from within nextcloud as well as from other sites? Do we need to create a proxy Controller without CORS?

blizzz:
Perhaps taking away CORS? But then, it was added for a reason. Or not?

marcelklehr:
Without CORS it works, so there's some mistake in how I added CORS, I guess.

marcelklehr:
Perhaps we need something like this: https://github.com/nextcloud/passman/blob/f8c13471b78b76a0576dd29e8c32d9e1e1f2359d/middleware/apimiddleware.php

marcelklehr:
BernhardPosselt on IRC pointed out to me as well that CORS drops sessions, so we do need two controllers. The question is if this pr makes sense at all in this case. The whole idea was to have only one API controller. In this PR I basically merged the two, but now we're exposing the same interface at two different routes... Might still be better than having two entirely different interfaces on the two endpoints... What do you say?

blizzz:
Sidenote: just came across the added ``@NoCSRFRequired`` annotations. Those would make it vulnerable to CSFR attacks. For pure API functionality this is necessary to drop indeed, but not within our interface, because sessions are active.

If there is no use case for CORS (i am not aware of an request?), we can leave that out. The Passman middlewere reference would be about it?

Two different routes are OK imho, however I do not see them duplicated? 

Actually, now I am thinking whether we should not keep a dedicated apicontroller, which would just inherit the internal one and just change the annotations, and call the parent. Probably there is a hook, though.

marcelklehr:
I agree on the CSRF issue.

CORS is not important to me, either.

CORS is important imo, since it's an API.

I agree on the approach and have started working on this.

marcelklehr:
I think @CORS imight be necessary for security, though, since NoCSRFRequired is added to API endpoints and all...?

blizzz:
> I think @CORS imight be necessary for security, though, since NoCSRFRequired is added to API endpoints and all...?

Not really for security. While it disables sessions (client would also need to reauthenticate), it also allows to be used across origins, i.e. from other websites/webapps. Sounds OK to me to include it nevertheless.

blizzz:
The changes seem to work so far from the UI. However, when returning bookmarks, only a piece of the data is returned back.

Expected:

spectacle hz7378

Returned:

spectacle yv7378

You'll see it as the date in the list is represented as "Invalid date" in the UI.

marcelklehr:
This is a symptom of merging the two APIs, it can be easily fixed by either modifying the client-side to included a parameter that tells the API all attributes it needs, or by modifying the API to deliver all those by default. Your call :)

blizzz:
Just return all of them. It keeps things simple, and the only extra cost comes from the tags. This is fine, and nothing can go wrong by chance (e.g
 updating a bookmark).

controller/rest/bookmarkcontroller.php
```diff
@@ -218,7 +276,7 @@ public function deleteBookmark($id = -1) {
 	}
 
 	/**
-	  @NoAdminRequired
+		@NoAdminRequired
```
blizzz: * missing

controller/rest/bookmarkcontroller.php
```diff
@@ -243,7 +301,7 @@ public function clickBookmark($url = "") {
...
```
blizzz: * missing

controller/rest/bookmarkcontroller.php
```diff
@@ -271,7 +329,7 @@ public function importBookmark() {
...
```
blizzz: * missing

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,151 @@
...
+namespace OCA\Bookmarks\Controller\Rest;
+
+use OCP\IDBConnection;
+use OCP\IL10N;
+use \OCP\IRequest;
+use \OCP\AppFramework\ApiController;
+use \OCP\AppFramework\Http\JSONResponse;
+use \OCP\AppFramework\Http;
+use \OC\User\Manager;
+use OCA\Bookmarks\Controller\Rest\BookmarkController;
```
blizzz: not in use

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,151 @@
...
+use \OCA\Bookmarks\Controller\Lib\Bookmarks;
+use \OCA\Bookmarks\Controller\Lib\ExportResponse;
```
blizzz: this and the next two are also not in use

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @return JSONResponse
+	 *
+	 * @NoAdminRequired
+	 */
+	public function getBookmarks(
+		$type = "bookmark",
+		$tag = '', // legacy
+		$page = 0,
+		$sort = "bookmarks_sorting_recent", // legacy
+		$user = null,
```
blizzz: this and the next parameters are missing in phpdoc

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @param array $item
+	 * @param string $title
+	 * @param bool $is_public
+	 * @param string $description
+	 * @return JSONResponse
+	 *
+	 * @NoAdminRequired
+	 */
+	public function newBookmark($url = "", $item = array(), $title = "", $is_public = false, $description = "") {
+		return $this->publicController->newBookmark($url, $item, $title, $is_publiic, $description);
```
blizzz: typo: ``$is_publiic`` has an i too much

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @return \OCP\AppFramework\Http\JSONResponse
+	 *
+	 * @NoAdminRequired
+	 */
+	public function deleteBookmark($id = -1) {
+		return $this->publicController->deleteBookmark($id);
+	}
+
+	/**
+	  @NoAdminRequired
```
blizzz: * missing (same with next two methods)

controller/rest/internaltagscontroller.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace OCA\Bookmarks\Controller\Rest;
+
+use \OCA\Bookmarks\Controller\Lib\Bookmarks;
+use \OCP\AppFramework\Http\JSONResponse;
+use \OCP\AppFramework\Http;
```
blizzz: not in use

controller/rest/internaltagscontroller.php
```diff
@@ -0,0 +1,48 @@
...
+use \OCP\AppFramework\ApiController;
+use OCA\Bookmarks\Controller\Rest\InternalTagsController;
```
blizzz: not in use

tests/bookmarkcontroller_test.php
```diff
@@ -19,91 +19,111 @@ class Test_PublicController_Bookmarks extends TestCase {
 	private $db;
 	private $userManager;
 	/** @var	PublicController */
```
blizzz: should become ``BookmarkController``

tests/bookmarkcontroller_test.php
```diff
@@ -19,91 +19,111 @@ class Test_PublicController_Bookmarks extends TestCase {
...
+	private $controller;
 	private $publicController;
 
 	protected function setUp() {
 		parent::setUp();
 
 		$this->userid = "testuser";
+		$this->otherUser = "otheruser";
```
blizzz: declaration missing

tests/bookmarkcontroller_test.php
```diff
@@ -19,91 +19,111 @@ class Test_PublicController_Bookmarks extends TestCase {
 	private $db;
 	private $userManager;
 	/** @var	PublicController */
+	private $controller;
 	private $publicController;
```
blizzz: would need the ``/** @var	BookmarkController */`` typehint, too

controller/rest/internalbookmarkcontroller.php
marcelklehr: I disagree. 12 lines below it is used ;)

blizzz: Okay, it is used, you are totally right. However, it is in the same namespace, so the use statement is not necessary still. :)

marcelklehr: Ah, I see. Will remove it.

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,153 @@
...
+use OCA\Bookmarks\Controller\Rest\BookmarkController;
+use \OCA\Bookmarks\Controller\Lib\Bookmarks;
+
+class InternalBookmarkController extends ApiController {
+
+	private $publicController;
+
+	public function __construct($appName, IRequest $request, $userId, IDBConnection $db, IL10N $l10n, Bookmarks $bookmarks, Manager $userManager) {
+		parent::__construct($appName, $request);
+	  $this->publicController = new BookmarkController($appname, $request, $userId, $db, $l10n, $bookmarks, $userManager);
```
blizzz: Oh, missed this before. Typo in ``$appname``, the n must be upper case.

controller/rest/bookmarkcontroller.php
```diff
@@ -57,27 +60,81 @@ public function legacyGetBookmarks($type = "bookmark", $tag = '', $page = 0, $so
 	 * @param string $tag
 	 * @param int $page
 	 * @param string $sort
+	 * @param string user
```
blizzz: sorry, but the $'s are missing in the var names

controller/rest/internalbookmarkcontroller.php
```diff
@@ -0,0 +1,152 @@
...
+		parent::__construct($appName, $request);
+	  $this->publicController = new BookmarkController($appName, $request, $userId, $db, $l10n, $bookmarks, $userManager);
+	}
+
+	/**
+	 * @param string $type
+	 * @param string $tag
+	 * @param int $page
+	 * @param string $sort
+	 * @param string user
```
blizzz: here too 😇

blizzz:
I leave the two cosmetics for a quick additional PR. And use the moment to merge this. Good work, thanks a lot @marcelklehr :) 🚀 🎸

PR #339, opened 2017-02-16
Add rel="nofollow noopener"

brantje:
The page we're linking to gains partial access to the linking page via the window.opener object.
For more info see this: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/


blizzz:
Nofollow should not really matter currently, since there is no public page to get bookmarks.  

Anyway, Ok, thx 👍

Issue #341, opened 2017-04-03
Bookmarks 0.9.1 throws error if used in Owncloud 9.1.4 because of missing theming support


e-alfred:
Hello,

Bookmarks 0.9.1 only shows a blank page if I try to access the app in Owncloud 9.1.4. Here is the error message in the log:

`
"Call to undefined method OC\\Server::getThemingDefaults() at \/var\/www\/owncloud\/apps\/bookmarks\/templates\/main.php#22
`

I fixed it by editing the bookmarklet() function in /templates/main.php:

```
function bookmarklet($bookmarkleturl) {
$l = \OC::$server->getL10N('bookmarks');
# $defaults = \OC::$server->getThemingDefaults();
$blet = "javascript:(function(){var a=window,b=document,c=encodeURIComponent,e=c(document.title),d=a.open('";
$blet .= $bookmarkleturl;
$blet .= "?output=popup&url='+c(b.location)+'&title='+e,'bkmk_popup','left='+((a.screenX||a.screenLeft)+10)+',top='+((a.screenY||a.screenTop)+10)+',height=400px,width=550px,resizable=1,alwaysRaised=1');a.setTimeout(function(){d.focus()},300);})();";
$help_msg = $l->t('Drag this to your browser bookmarks and click it, when you want to bookmark a webpage quickly:');
$output = '<div id="bookmarklet_hint" class="bkm_hint">' . $help_msg . '</div><a class="button bookmarklet" href="' . $blet . '"> Add to Owncloud</a>';
return $output;
}
?>
```

blizzz:
Thanks for your report! 

As far as I know there is no bookmarks dev using ownCloud. So far we were driving the usual tests against it and see how it went.  Apparently this part was uncovered.

As such pull requests fixing this while staying compatible are welcome.

PR #342, opened 2017-04-06
Adjust tests for phpunit

blizzz:
Since they started to fail…

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: php
 php:
-  - 5.5
   - 5.6
   - 7
   - 7.1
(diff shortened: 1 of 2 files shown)


marcelklehr:
Mh, things are failing that didn't fail before, right? Weird.

blizzz:
Exactly. But now we are down to failing rename tests again, which is expected \o/ Plus, failing OC tests, because their master is 10. Since we got an error report in this regard (https://github.com/nextcloud/bookmarks/issues/341) and an [extremely hostile post by assumingly on of their officials](https://central.owncloud.org/t/nextcloud-bookmark-app-calles-undefined-methode/5724/2) I am pretty open for dropping this support at all, at least of no contributions comes from that side. Anyway, with the latest commit I increased max-version for oc for now, let's see how these tests go.

blizzz:
For one it was PHP 5.5 running out of life that caused fails, and I also took over https://github.com/nextcloud/activity/pull/126 from the activity app.

blizzz:
Now there is just one unexpected one missing, being OC 9.1 against PHP 7/sqlite. Seems that goes with a newer phpunit version that tries to load a file which is not there anymore.

blizzz:
I put the one failing OC test as allowed (for now).

PR #360, opened 2017-05-15
Fix lib#findUniqueBookmark: Return all tags

### LINKED ISSUE #357: REST: POST new Bookmark only first set tag will be returned via tags
Hello,

so I've tried the sollution you posted here https://github.com/nextcloud/bookmarks/issues/353, and yes it worked. However the json that will be returned by sending such a query now contains only the first tag I've set.

### Example

POST `http://msi-7816:8080/index.php/apps/bookmarks/public/rest/v2/bookmark?url=https://fachschaft.in&item[tags][]=bla&item[tags][]=fuddeldud`

```
{
  "item": {
    "id": "9",
    "url": "https://fachschaft.in",
    "title": "News",
    "user_id": "admin",
    "description": "",
    "public": "0",
    "added": "1494838251",
    "lastmodified": "1494838251",
    "clickcount": "0",
    "tags": "bla"
  },
  "status": "success"
}
```

But the tags get set anyway. So that works

marcelklehr:
not just the first one :)

fixes #357

### DIFF
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -115,7 +115,7 @@ class Bookmarks {
 			->select('tag')
 			->from('bookmarks_tags')
 			->where($qb->expr()->eq('bookmark_id', $qb->createNamedParameter($id)));
-		$result['tags'] = $qb->execute()->fetchColumn();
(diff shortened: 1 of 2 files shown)


blizzz:
Perhaps in the test you can also insert two or more tags and check their retrieval?

PR #362, opened 2017-05-15
README: Fix API docs

marcelklehr:
Omit a param -> resets attr

My bad.


blizzz:
… although, perhaps we should change the behaviour for values to be explicitly set to be removed?

marcelklehr:
Yep, the point was to not break compatibility with the internal API, but since its usage entails always sending all attrs, we could probably safely add in some array_merge step.

marcelklehr:
closing in favor of #363

PR #363, opened 2017-05-16
API: Allow omitting params in editBookmark

### LINKED ISSUE #359: REST: Not able to edit bookmarks
I'm trying:

PUT `http://msi-7816:8080/index.php/apps/bookmarks/public/rest/v2/bookmark/15?titel=asdf`

I'm getting a `400`

Same with

PUT `http://msi-7816:8080/index.php/apps/bookmarks/public/rest/v2/bookmark/15?record_id=15&titel=asdf`

marcelklehr:
This is preliminary. I'm not sure it'll work. Especially whether it allows for actually removing/resetting attribs.

Fixes #359

### DIFF
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -229,23 +229,29 @@ class BookmarkController extends ApiController {
 	 * @NoCSRFRequired
 	 * @CORS
 	 */
-	public function editBookmark($id = null, $url = "", $item = array(), $title = "", $is_public = false, $record_id = null, $description = "") {
+	public function editBookmark($id = null, $url = null, $item = null, $title = null, $is_public = null, $record_id = null, $description = null) {
 
-		// Check if it is a valid URL
-		$urlData = parse_url($url);
-		if(!$this->isProperURL($urlData)) {
+		if ($record_id !== null) {
+			$id = $record_id;
+		}
+		
+		$bookmark = $this->bookmarks-findUniqueBookmark($id, $this->userId);
+		$bookmark = array_merge($bookmark, [
+			'id' => $id,
+			'url' => $url,
+			'tags' => isset($item) ? $item['tags'] : null,
+			'title' => $title,
+			'is_public' => $is_public,
+			'description' => $description
+		]);
+
+		// Check if url and id are valid
+		$urlData = parse_url($bookmark['url']);
+		if(!$this->isProperURL($urlData) || !is_numeric($bookmark['id'])) {
 			return new JSONResponse(array(), Http::STATUS_BAD_REQUEST);
 		}
 
-		if ($record_id == null) {
-			return new JSONResponse(array(), Http::STATUS_BAD_REQUEST);
-		}
-
-		$tags = isset($item['tags']) ? $item['tags'] : array();
-
-		if (is_numeric($record_id)) {
-			$id = $this->bookmarks->editBookmark($this->userId, $record_id, $url, $title, $tags, $description, $is_public = false);
-		}
+		$id = $this->bookmarks->editBookmark($this->userId, $bookmark['id'], $bookmark['url'], $bookmark['title'], $bookmark['tags'], $bookmark['description'], $bookmark['is_public']);
 
 		$bm = $this->bookmarks->findUniqueBookmark($id, $this->userId);
 		return new JSONResponse(array('item' => $bm, 'status' => 'success'));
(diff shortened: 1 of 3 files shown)


marcelklehr:
Finally works! Travis doesn't like nc 12, though, ofc

blizzz:
corner case: cannot remove the last remaining tag

blizzz:
… or in other words: it is not possible to remove all tags

marcelklehr:
yea, how do you think this should be done?

blizzz:
If ``item[tags]`` is set, but to an empty array => remove all tags, however if it is not set, keep them.

Actually this works with the REST API :) The Web UI does not.

blizzz:
The REST call failed, because i did not provide other optional fields like url. they' be overwritten by null.

blizzz:
@marcelklehr you can incorporate following to have all incoming properties optional:

```
index 50893da1..a118f861 100644
@@ -237,14 +237,19 @@ class BookmarkController extends ApiController {
        
        $bookmark = $this->bookmarks->findUniqueBookmark($id, $this->userId);
        $newProps = [
-           'id' => $id,
            'url' => $url,
            'title' => $title,
            'is_public' => $is_public,
            'description' => $description
        ];
-       if (isset($item) && isset($item['tags'])) $newProps['tags'] = $item['tags'];
-       $bookmark = array_merge($bookmark, $newProps);
+       foreach ($newProps as $prop => $value) {
+           if(!is_null($value)) {
+               $bookmark[$prop] = $value;
+           }
+       }
+       if(is_array($item) && isset($item['tags']) && is_array($item['tags'])) {
+           $bookmark['tags'] = $item['tags'];
+       }
 
        // Check if url and id are valid
        $urlData = parse_url($bookmark['url']);
```

It should not be possible to change the ID.

For the webUI part some adjustments are needed.


Anyway
``curl -X PUT -u "user:password" -d "title=Foobar" 'http://my.server/index.php/apps/bookmarks/public/rest/v2/bookmark/3'`` keeps the Tags while ``curl -X PUT -u "user:password" -d "title=Barfoo&item[tags][]=" 'http://my.server/index.php/apps/bookmarks/public/rest/v2/bookmark/3'`` deletes the tags

blizzz:
And for the web UI you can apply

```diff
diff --git a/js/bookmarks.js b/js/bookmarks.js
index 0e27a9aa..8a4be22c 100644
--- a/js/bookmarks.js
+++ b/js/bookmarks.js
@@ -279,6 +279,11 @@ function editBookmark() {
    rec_form.bind('submit', function (event) {
        event.preventDefault();
        var form_values = $(this).serialize();
+       if(form_values.indexOf('item%5Btags%5D') === -1) {
+           // if not tag is selected, the input field does not exist, so
+           // we need to manually insert an empty array
+           form_values += "&item%5Btags%5D%5B%5D=";
+       }
        $.ajax({
            type: 'PUT',
            url: $(this).attr('action') + "/" + this.elements['record_id'].value,
```

marcelklehr:
I think the id param can only be passed via the path, not via query string, can it? In any case it's not really possible to change the id, as it's also used to select the bookmark to edit, I think...

Thanks for the client-side fix, though! :)

blizzz:
@marcelklehr the id change is a side-catch (did not try it via an additional parameter, probabl won't work, nevertheless).  The main thing was properly handling the incoming item[tags] parameter.

marcelklehr:
Is it ok, if I leave in the array_merge? Feels cleaner.

blizzz:
> Is it ok, if I leave in the array_merge? Feels cleaner.

It'll override existing values with null, if the parameter was not specified. → Should be covered with unit tests.

marcelklehr:
Bah, I hate this distinction between isset and null... *shakes head*

controller/rest/bookmarkcontroller.php
```diff
@@ -229,24 +229,40 @@ public function legacyEditBookmark($id = null, $url = "", $item = array(), $titl
...
 	 */
-	public function editBookmark($id = null, $url = "", $item = array(), $title = "", $is_public = false, $record_id = null, $description = "") {
+	public function editBookmark($id = null, $url = null, $item = null, $title = null, $is_public = null, $record_id = null, $description = null) {
 
-		// Check if it is a valid URL
-		$urlData = parse_url($url);
-		if(!$this->isProperURL($urlData)) {
-			return new JSONResponse(array(), Http::STATUS_BAD_REQUEST);
+		if ($record_id !== null) {
+			$id = $record_id;
```
blizzz: Shouldn't we just bury $record_id? The ID to operate against is provided by the $id attribute. And should not be overridden imho, this way you could overwrite the ID with a foreign way. Which brings me to the thought, to check whether we properly check ownership of the bookmark in any method in general… OK, for another PR.

Issue #366, opened 2017-05-22
Filtering by tags throws InvalidFieldNameException


pixi-p:
When selecting more than 1 tag the bookmarks list display gets corrupted as shown at the screenshot below:
![bookmarks_issue](https://cloud.githubusercontent.com/assets/28874613/26326030/c36148a4-3f39-11e7-840b-bd09be0f8910.png)

If multiple tags are selected all bookmarks are returned having any of the selected tags assigned instead of bookmarks having all selected tags assigned.

The issue seems to be with tag's name being surrounded with backticks characters in SQL query.

### Expected behaviour
- No exception logged in the log
- The tags assigned to the bookmark are visible only once
- Only bookmarks having all selected tags assigned are displayed

### Actual behaviour
Nextcloud shows the following log entry:
```
Doctrine\DBAL\Exception\InvalidFieldNameException: An exception occurred while executing 'SELECT t.`tag`, COUNT(t.`bookmark_id`) AS `nbr` FROM `oc_bookmarks_tags` t INNER JOIN `oc_bookmarks` b ON (b.`id` = t.`bookmark_id`) AND (b.`user_id` = ?) WHERE t.`tag` NOT IN (`dotNET`) GROUP BY t.`tag` ORDER BY `nbr` DESC LIMIT 18446744073709551615 OFFSET 0' with params ["pixi"]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'dotNET' in 'where clause'
```

### Server configuration

**Operating system**: Ubuntu 16.04 LTS

**Web server:** apache 2.4.18-2ubuntu3.2

**Database:** mysql  5.7.18-0ubuntu0.16.04.1

**PHP version:** 7.0.15-0ubuntu0.16.04.4

**Nextcloud version:** 11.0.3

**Bookmarks version:** 0.10.0

**Updated from an older Nextcloud or fresh install:**

**Signing status:**

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
  - activity: 2.4.1
  - bookmarks: 0.10.0
  - calendar: 1.5.3
  - comments: 1.1.0
  - contacts: 1.5.3
  - dav: 1.1.1
  - direct_menu: 0.10.2
  - external: true
  - federatedfilesharing: 1.1.1
  - federation: 1.1.1
  - files: 1.6.1
  - files_automatedtagging: 1.1.1
  - files_external: 1.1.2
  - files_markdown: 1.0.1
[...]

PR #367, opened 2017-05-22
Fix findTags: Use query param to inject tags

marcelklehr:
Fixed the exception thrown in #366

### DIFF
--- a/tests/lib_bookmark_test.php
+++ b/tests/lib_bookmark_test.php
@@ -93,8 +93,8 @@ class Test_LibBookmarks_Bookmarks extends TestCase {
 		$this->libBookmarks->addBookmark($this->userid, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
 		$this->libBookmarks->addBookmark($this->userid, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
 		
-		$findTags = $this->libBookmarks->findTags($this->userid, ["two"]);
-		$this->assertEquals([array('tag' => 'one', 'nbr' => 2), array('tag' => 'four', 'nbr' => 1)], $findTags);
+		$findTags = $this->libBookmarks->findTags($this->userid, ["two", "one"]);
+		$this->assertEquals([array('tag' => 'four', 'nbr' => 1)], $findTags);
 	}
   
 	function testRenameTag() {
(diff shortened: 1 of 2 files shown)


marcelklehr:
nc master is failing because of version spec in appinfo.

blizzz:
>nc master is failing because of version spec in appinfo.
 
let me fix that

blizzz:
can we have a test for this?

blizzz:
Needs a rebase, tests are failing

marcelklehr:
rebased.

blizzz:
Still failing:

```
There was 1 failure:
1) OCA\Bookmarks\Tests\Test_LibBookmarks_Bookmarks::testFindTagsFilter
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
-        'tag' => 'four'
-        'nbr' => 1
+        'tag' => 'one'
+        'nbr' => '2'
     )
     1 => Array (
-        'tag' => 'one'
-        'nbr' => 2
+        'tag' => 'four'
+        'nbr' => '1'
     )
-    2 => Array (...)
 )

/home/travis/build/nextcloud/server/apps/bookmarks/tests/lib_bookmark_test.php:97
```

blizzz:
I can still reproduce the same issue:

spectacle z13693

It also adds the tags "two", "three", "four" which are not used with the currently stored bookmarks. They somehow got stuck in the DB… perhaps not critical.

tests/lib_bookmark_test.php
```diff
@@ -85,6 +85,17 @@ function testFindTags() {
...
 	}
+	
+	function testFindTagsFilter() {
+		$this->cleanDB();
+		$this->assertEquals($this->libBookmarks->findTags($this->userid), array());
+		$this->libBookmarks->addBookmark($this->userid, "http://www.google.de", "Google", array("one"), "PrivateNoTag", false);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.heise.de", "Heise", array("one", "two"), "PrivatTag", false);
+		$this->libBookmarks->addBookmark($this->userid, "http://www.golem.de", "Golem", array("four"), "PublicNoTag", true);
+		
+		$findTags = $this->libBookmarks->findTags($this->userid, ["two"]);
```
blizzz: you need to test with filtering two tags and assert that no tag is used multiple times and no wrong match is pulled it (cf. screenshot in my other comment).

marcelklehr:
Ok. Testing with two tag filters, now. This doesn't fix the tags appear multiple times in the bookmark view, btw. It merely fixes the exception ;)

blizzz:
It should fail then, the assertion is not strict enough

marcelklehr:
I don't understand what you mean :?

marcelklehr:
It shouldn't fail because tags appearing multiple times is not a problem of the find Tags method.

blizzz:
D'oh!

PR #369, opened 2017-05-22
Fix rename tag

marcelklehr:
duplicates handling handled one duplicate only


marcelklehr:
Composer is acting out on travis, I think.. :/

marcelklehr:
The DELETE query removes to-be duplicates before renaming. ;)

marcelklehr:
(rebased.)

PR #374, opened 2017-05-25
makes add_url more wide for see the +-picture

blackcrack:
makes add_url more wide for see the +-picture
and use the space where it is easy to big
and makes more fit.


blizzz:
very good, thank you :)

blackcrack:
*thumps up* you're welcome, certainly, with pleasure happen

PR #377, opened 2017-05-27
Fix findBookmarks: Return empty tags if none found

### LINKED ISSUE #371: REST: Bookmark with no tags contains empty tag
Bookmarks that do not have a tag now return an empty "" tag when trying to tags:
```
"tags": [
  ""
]
```

### Call
GET `http://msi-7816:8080/index.php/apps/bookmarks/public/rest/v2/bookmark`

### Replay
```
{
  "data": [
    {
      "id": "17",
      "url": "http://asdf.com",
      "title": "asdf",
      "user_id": "admin",
      "description": "",
      "public": "0",
      "added": "1495714580",
      "lastmodified": "1495714580",
      "clickcount": "0",
      "tags": [
        ""
      ]
    }
  ]
}
```

marcelklehr:
fixes #371

### DIFF
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -226,7 +226,12 @@ class Bookmarks {
 		$bookmarks = array();
 		foreach ($results as $result) {
 			if ($returnTags) {
-				$result['tags'] = is_null($result['tags']) ? [] : explode(',', $result['tags']);
+				// pgsql returns "", others null
+				if($result['tags'] === null || $result['tags'] === '') {
+					$result['tags'] = [];
+				} else {
+					$result['tags'] = explode(',', $result['tags']);
+				}
 			} else {
 				unset($result['tags']);
 			}
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -226,7 +226,11 @@ class Bookmarks {
 		$bookmarks = array();
 		foreach ($results as $result) {
 			if ($returnTags) {
-				$result['tags'] = explode(',', $result['tags']);
+				if ($result['tags'] != "") {
+					$result['tags'] = explode(',', $result['tags']);
+				} else {
+					$result['tags'] = [];
+				}
 			} else {
 				unset($result['tags']);
 			}
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -226,11 +226,7 @@ class Bookmarks {
 		$bookmarks = array();
 		foreach ($results as $result) {
 			if ($returnTags) {
-				if ($result['tags'] != "") {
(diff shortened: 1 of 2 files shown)


controller/lib/bookmarks.php
```diff
@@ -226,7 +226,11 @@ public function findBookmarks(
 		$bookmarks = array();
 		foreach ($results as $result) {
 			if ($returnTags) {
-				$result['tags'] = explode(',', $result['tags']);
+				if ($result['tags'] != "") {
```
blizzz: please always use strict comparison (e.g. ``!==``). This won't work then, because ``$result['tags']`` is null. Also, not formatting is not enough, because an array is expected.

The one liner can be kept, but with a condition:

```php
$result['tags'] = is_null($result['tags']) ? [] : explode(',', $result['tags']);
```

or the longer version with full if-condition:

```php
if($result['tags'] === null) {
	$result['tags'] = [];
} else {
	$result['tags'] = explode(',', $result['tags']);
}
```

blizzz (dismissed):

blizzz:
ah, let me push directly :)

blizzz:
@marcelklehr pushed my changes. If you agree and tests run, please merge.

controller/lib/bookmarks.php
```diff
@@ -226,7 +226,7 @@ public function findBookmarks(
 		$bookmarks = array();
 		foreach ($results as $result) {
 			if ($returnTags) {
-				$result['tags'] = explode(',', $result['tags']);
+				$result['tags'] = is_null($result['tags']) ? [] : explode(',', $result['tags']);
```
marcelklehr: `$result['tags']` is never null, though. See line 195: `$qb->createFunction("array_to_string(array_agg(" . $qb->getColumnName('t.tag') . "), ',')"), 'tags'...`

This should be `$result['tags'] === '' ?`, imo, don't you think?

marcelklehr (commented):
Change is regression

controller/lib/bookmarks.php
marcelklehr: $result['tags'] is not an array, that's why this loop exists in the first place ;)

controller/lib/bookmarks.php
blizzz: I went with the debugger through it, and it was clearly a null. However, Postgres tests were failing exactly here. Do you work against Postgres?

i'll have another look anyhow, thx!

blizzz: Two lines further you have the functions used for the other RDBMs, so it seems to be DB-realted indeed. I included now an empty string in the condition and turned it into the ``if``, because it got too long.

marcelklehr:
Thanks for clearing that up :) Good to go!

PR #408, opened 2017-11-19
New REST endpoint: read single bookmark

marcelklehr:
So, in order to add a tag without deleting the already existing ones, you need to know what tags are already set. Unfortunately there's no way to get that info on a single bookmark without fetching them all at once (uff.), ... until now :)

This pr adds a REST endpoint that proxies libBookmarks#findUniqueBookmark and even takes care of public/private bookmarks.

### DIFF
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -78,7 +78,7 @@ class BookmarkController extends ApiController {
 			}
 		}
 		$bm = $this->bookmarks->findUniqueBookmark($id, $user);
-		if ($public_only === TRUE && $bm['public'] !== '1') {
+		if ($public_only === TRUE && $bm['public'] != '1') {
 			$error = "Insufficient permissions";
 			return new JSONResponse(array('status' => 'error', 'data' => $error), Http::STATUS_BAD_REQUEST);
     }
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -269,7 +269,7 @@ class BookmarkController extends ApiController {
 		$newProps = [
 			'url' => $url,
 			'title' => $title,
-			'is_public' => $is_public,
+			'public' => $is_public,
 			'description' => $description
 		];
 		if (is_array($item) && isset($item['tags']) && is_array($item['tags'])) {
@@ -291,7 +291,7 @@ class BookmarkController extends ApiController {
 			$bookmark['tags'] = [];
 		}
 
-		$id = $this->bookmarks->editBookmark($this->userId, $bookmark['id'], $bookmark['url'], $bookmark['title'], $bookmark['tags'], $bookmark['description'], $bookmark['is_public']);
+		$id = $this->bookmarks->editBookmark($this->userId, $bookmark['id'], $bookmark['url'], $bookmark['title'], $bookmark['tags'], $bookmark['description'], $bookmark['public']);
 
 		$bm = $this->bookmarks->findUniqueBookmark($id, $this->userId);
 		return new JSONResponse(array('item' => $bm, 'status' => 'success'));
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -65,21 +65,21 @@ class BookmarkController extends ApiController {
 	 */
 	public function getSingleBookmark(
 		$id,
-    $user = null
+		$user = null
 	) {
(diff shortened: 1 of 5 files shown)


marcelklehr:
For the record. The tests fail on an unrelated technicality, of which I'm not sure if it's happening on master, too. (here we go again...)

blizzz:
Sorry for the delay. 

The tests fail all for postgre, other tests against master run fine

controller/rest/bookmarkcontroller.php
```diff
@@ -54,6 +54,36 @@ public function __construct($appName, IRequest $request, $userId, IDBConnection
...
+			$publicOnly = false;
+		}else {
+			$publicOnly = true;
+			if ($this->userManager->userExists($user) == false) {
+				$error = "User could not be identified";
+				return new JSONResponse(array('status' => 'error', 'data'=> $error), Http::STATUS_BAD_REQUEST);
+			}
+		}
+		$bm = $this->bookmarks->findUniqueBookmark($id, $user);
+    if ($publicOnly === TRUE && $bm['public'] !== '1') {
```
blizzz: the indentation here and in the function declaration is off

blizzz:
There's indeed not much potential what could go wrong… perhaps `$bm['public']` ends up as int?

marcelklehr:
re my comment before: I made that remark in reference to an earlier state of this pr :)

Seems now as if it's a type incoherence, yep.

marcelklehr:
Here we go :)

marcelklehr:
having this in the next release would allow floccus to sync folder structures by applying a tag with the folder path as an intermediate solution.

controller/rest/bookmarkcontroller.php
```diff
@@ -65,21 +65,21 @@ public function legacyGetBookmarks($type = "bookmark", $tag = '', $page = 0, $so
 	 */
 	public function getSingleBookmark(
 		$id,
-    $user = null
+		$user = null
 	) {
 		if ($user === null) {
 			$user = $this->userId;
-			$publicOnly = false;
+			$public_only = false;
```
blizzz: no, please don't, camelCase is the coding style https://docs.nextcloud.com/server/12/developer_manual/general/codingguidelines.html#objects-functions-arrays-variables

blizzz:
Let's return 404 for a non-existing bookmark:

```diff
diff --git a/controller/rest/bookmarkcontroller.php b/controller/rest/bookmarkcontroller.php
index bd65bd74..bf0e95b6 100644
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -78,7 +78,10 @@ class BookmarkController extends ApiController {
            }
        }
        $bm = $this->bookmarks->findUniqueBookmark($id, $user);
-       if ($public_only === TRUE && $bm['public'] != '1') {
+       if(!isset($bm['id'])) {
+           return new JSONResponse(['status' => 'error'], Http::STATUS_NOT_FOUND);
+       }
+       if ($public_only === true && isset($bm['public']) && $bm['public'] != '1') {
            $error = "Insufficient permissions";
            return new JSONResponse(array('status' => 'error', 'data' => $error), Http::STATUS_BAD_REQUEST);
     }
diff --git a/tests/bookmarkcontroller_test.php b/tests/bookmarkcontroller_test.php
index 3eabbda7..a88000a1 100644
--- a/tests/bookmarkcontroller_test.php
+++ b/tests/bookmarkcontroller_test.php
@@ -65,6 +65,15 @@ class Test_BookmarkController extends TestCase {
        $this->assertEquals("http://www.9gag.com", $data['item']['url']);
    }
 
+   function testPrivateReadNotFound() {
+       $this->cleanDB();
+       $this->setupBookmarks();
+       $output = $this->controller->getSingleBookmark(987);
+       $data = $output->getData();
+       $this->assertSame('error', $data['status']);
+       $this->assertSame(404, $output->getStatus());
+   }
+
   function testPublicReadSuccess() {
        $this->cleanDB();
        $this->setupBookmarks();

```

and get it in.

controller/rest/bookmarkcontroller.php
marcelklehr: that's what I thought, until I misinterpreted your comment: "keep true lower case"... :laughing:

blizzz (approved):

blizzz:
thx @marcelklehr :)

marcelklehr:
:tada:

PR #413, opened 2017-11-25
wider title, look normal on desktop

ghost:
title. seems like desktop / tablet would be default view. if want phone do media query or something


marcelklehr:
Looking at the alignment and the numbers in the css file, the intention seems to have been to have a table-like layout. But somehow using a P-tag for the tags introduced a new line which renders the carefully constructed columns pointless.

I suppose it makes sense to just embrace the P-tag give both the tags and the title their own line, for now. Until the new UI is ready.

marcelklehr:
thanks @kylepwarren! :)

PR #414, opened 2017-11-29
Update Readme

marcelklehr:


blizzz:
imho a bit of a negative touch, but perhaps it's just me ;)

marcelklehr:
You're right. I guess it does have a negative touch :D
Let me try to muster some positive energy and whip up something new :)

PR #415, opened 2017-11-30
Bookmarksperms

ghost:
Yeah, probably was unfinished table layout or something. 
Actually, was just inspecting it some more, looks like the description was supposed to go to the right of the title. I never filled those [descriptions] out. 

I guess who ever did that html wanted it to be able to collapse (for phones), that's probably why it's not a table. 

The date, on the right, should be pos. abs. the other stuff on the left(title description, tags) should not be absolute.

... There: description under the title. tested w/ long description, no description, narrow view port.

### DIFF
--- a/css/bookmarks.css
+++ b/css/bookmarks.css
@@ -292,17 +292,12 @@ ul.tagit li.tagit-choice .close .text-icon {
 }
 
 .bookmark_desc{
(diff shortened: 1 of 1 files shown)


marcelklehr:
Ah, makes sense.

PR #417, opened 2017-12-30
fixes #416

### LINKED ISSUE #416: Export bookmarks doesn't work
### Steps to reproduce
1. click on gear
2. click export

### Expected behaviour
An htm file should be downloaded

### Actual behaviour
No file gets downloaded . This error message instead

> File not found
The specified document has not been found on the server.
You can click here to return to Nextcloud.

### Server configuration

**Operating system**:
debian stable

**Web server:**
nginx

**Database:**
mysql

**PHP version:**
7x

**Nextcloud version:** (see Nextcloud admin page)
12.0.4

**Bookmarks version:** (see Nextcloud apps page)
0.10.1  (v9.1.0RC4-314-g3be4e33a)
**Updated from an older Nextcloud or fresh install:**
fresh

### Note
I've already diagnosed the issue and made a fix. I'll submit PR right after this ticket

damko:
the route
    array('name' => 'internal_bookmark#edit_bookmark', 'url' => '/bookmark/{id}', 'verb' => 'PUT'),
intercepts all the calls having an additional URL segment like /bookmark/export therefore editBookmark receives "export" as input and obviously returns failure without hitting the right controller method.

As I fix I moved down the routes having the parameter {id} so that the other routes like /bookmark/export or /bookmark/click can match before processing the routes with parameter.

I'm not sure this is the best fix. Another way, out of the top of my head, could be to rename the endpoint as `export_bookmarks` but this would break old clients.

---
Additionally I made sure that the exportBookmark endpoint can be called via public API (see method's header)


js/bookmarks.js
```diff
@@ -312,7 +312,7 @@ function updateBookmarksList(bookmark, position) {
 	bookmark = $.extend({title: '', description: '', added_date: new Date('now'), tags: []}, bookmark);
 	var tags = bookmark.tags;
 	var taglist = '';
-	for (var i = 0, len = tags.length; i < len; ++i) {
```
marcelklehr: this change is a noopo, no?

css/bookmarks.css
```diff
@@ -215,16 +215,25 @@ li.tag_list {
     overflow: hidden;
 }
 
+li.tag_list > li {
+    max-height: 30px;
+    display: inline-block;
+}
+
 li.tag_list a {
```
marcelklehr: these css changes should go into a separate pr.

marcelklehr (changes requested):
Other than the comments, looks good.

controller/lib/bookmarks.php
```diff
@@ -689,18 +689,17 @@ public function getURLMetadata($url, $tryHarder = false) {
...
-		$filterTag = array();
-		foreach ($tags as $tag) {
-			if (trim($tag) != '')
-				$filterTag[] = trim($tag);
+		if(is_array($line)) {
+			array_filter(array_map('trim',$line),'strlen');
+			return $line;
 		}
-		return $filterTag;
+		if(!is_string($line)) return array();
```
blizzz: please always  use curly brackets

marcelklehr:
@damko Can you update this pr?

blizzz:
is continued in #421  and #422 → closing

PR #420, opened 2018-01-10
UI: Don't escape URL for editing a bookmark

### LINKED ISSUE #401: URLs with special characters
Hello,

Seems to be a little problem with URLs containing special characters when editing a bookmark.

Example :
I added a bookmark for this page : **https://fr.wikipedia.org/wiki/Correspondance_entre_logiciels_libres_et_logiciels_propriétaires** which contains the letter 'é'.
This works fine.

But, if I edit this bookmark, the 'é' is converted and the URL reads now : **https://fr.wikipedia.org/wiki/Correspondance_entre_logiciels_libres_et_logiciels_propri%C3%A9taires**
This occurs as soon as I click on the Edit button.

Clicking on the modified bookmark generates this URL : **https://fr.wikipedia.org/wiki/Correspondance_entre_logiciels_libres_et_logiciels_propri%25C3%25A9taires**
**25** is added after the % sign each time the bookmark is edited...

Did I miss something ?

I use version 0.10.1 on Nextcloud 12.0.2.

Thanks a lot !

marcelklehr:
fixes #401

Actually, the url is still escaped, but not for use as a url, but for use as an HTML attribute value...

### DIFF
--- a/templates/js_tpl.php
+++ b/templates/js_tpl.php
@@ -29,7 +29,7 @@
             </p>
             <p class="bookmark_form_url">
                 <input type="text" name="url" placeholder="<?php p($l->t('The address of the page')); ?>"
-                       value="<&= encodeURI(url)&>"/>
+                       value="<&= escapeHTML(url)&>"/>
             </p>
             <div class="bookmark_form_tags"><ul>
                     <& for ( var i = 0; i < tags.length; i++ ) { &>


blizzz (changes requested):
The edit check does not work anymore: if you bookmark the site the second time the bookmarklet should show a warning that you are editing an existing bookmark. But it is not happening, actually the second attempt is totally lost. I don't even get log output. network console reports a 200.

marcelklehr:
mmh. are you sure that is new?

marcelklehr:
I cannot reproduce this problem with neither this branch nor master... see #429

blizzz:
@marcelklehr i have the same issue on master with https://fr.wikipedia.org/wiki/Correspondance_entre_logiciels_libres_et_logiciels_propri%C3%A9taires, but just with ascii-domains it works. So, a bug, but not regression.

blizzz (approved):

PR #421, opened 2018-01-10
Fix export

### LINKED ISSUE #416: Export bookmarks doesn't work

marcelklehr:
fixes #416 
(cherry picked from #417)


controller/rest/bookmarkcontroller.php
```diff
@@ -396,9 +393,14 @@ public function importBookmark() {
...
 
 	/**
-	 * 
+	 * Hit this GET endpoint to export bookmarks via your API client.
+	 * http://server_ip/nextcloud/index.php/apps/bookmarks/public/rest/v2/bookmark/export
+	 * Basic authentication required.
 	 * @return \OCP\AppFramework\Http\Response
+	 *
 	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
blizzz: that makes sense, but we should use the internal one, shouldn't we?

marcelklehr: I think the idea is to allow 3rd-party clients to export stuff as well (see https://github.com/nextcloud/bookmarks/pull/421/files#diff-0e033fd8797499a87a6bf2ffccef5c3cR43 where this is actually already in place)

blizzz (approved):

Issue #5, opened 2013-04-10
Add Sharing / Use share api [$145]


eMerzh:
The Current version of bookmark does not allow to share bookmark using the sharing api.

Let the user share a tag with users and using a "private link".

Export a simple page with the bookmarks and an feed (atom, rss, ..) .

pixi:
Any news on the sharing?

markus61:
+1

shaneonabike:
I definitely think this would be really useful and practical to have the ability to share all your bookmarks or a few with other users like calendars

BugHunter2k:
Any News? I'd like to see this too.

ronnystandtke:
Another +1. We would like to share a collection of bookmarks for a project with external people (they have no user account on our ownCloud instance). We already share some files with these external people via a public link. It would be cool if we could just integrate our bookmarks into this shared file folder as clickable links (maybe even visually separated from the ordinary shared files and folders). This way we only have to share one link and everything that is needed for our project is easily accessible for our external people.

brozkeff:
+1
Our company is looking exactly for this functionality, lack of sharing in current bookmarks implementation means it is useless for us at this moment.

pascaldaniela:
+1
exactly what we're looking for to avoid having yet another service to add in our organization. Could we have any feedback from an owncloud developer?
thanks!

xila76:
:+1:

ChrisEastGaming:
Hey! You can get free, steam games. If you can get please click the link. It's very important for me. No virus! 100% work! Just one click. Please.
 https://steamgifts.co/go/?id=RvqdcyTKd

TheNomad11:
Yes, please, I actually expected this feature when I installed it :)

marcelklehr:
:+1: :)

Endogen:
Any news on this? I would really love to have this feature

marcelklehr:
This is the status quo on this feature:
Currently, there is a database field for each bookmark reserved to hold information whether it is public or private. The API can handle the field and even querying it is supported, so you can for example retrieve the public bookmarks of a specific user.

So going forward, we'd have to discuss the exact concept for sharing, specifically whether a private/public mode is sufficient (I've never used the sharing API myself, so I'd have to do some reading on it). Also, some concept would need to be developed for how sharing should work in the ui.

I'm open to implementing this, but the ETA is unknown as of yet.

ccoenen:
[...]

PR #424, opened 2018-01-12
new ui

### LINKED ISSUE #12: No refresh when tagging bookmark / "no tagged" filter
Hi,

Firstly I would like to congratulate all the contributors of OC. This project is awesome ;)

I have a lot of bookmarks (100 ~ 150)...
I started to tag them but I had to stop because each update refreshes the page and displays the first bookmarks. Then I have to scroll down and search the bookmarks without tags, it's very slow and frustating...

My proposed changes are :
- Save changes for every update of bookmarks without refreshing,
  or
- Add a filter "without tag(s)" to update even refresh is still active.

Thanks for reading ;)

### LINKED ISSUE #55: grid view instead of list view for bookmarks
We are working on a grid view for the bookmarks category instead of a list view, because we think a that will make it more simple and attractive to browse the bookmarks. Our idea is to show the favicon or screenshot, titel, url, short description within every rectangle and showing the editing tools by hovering over them. In our opinion this will make the use of the bookmark tool more efficient than a list view. The list view wastes useful space. Do you have any comments on that issue? cc @laburramuge cc @flause

### LINKED ISSUE #80: bookmarks doesn't give use failure states on import
the bookmarks module fails silently if the import doesn't work for some reason.

Expectation: On failure to import user should know why.

### LINKED ISSUE #131: Bookmarks search not working in OC8
Hi, when I go to my bookmarks app and in the top right corner of the screen I try to search, in owncloud logs this appears:

```
{"reqId":"7f27997e0d1a4269fbcad40e11c207a2",
"remoteAddr":"190.31.13.117",
"app":"index",
"message":"Exception: {\"Message\":\"An exception occurred while executing 'SELECT * FROM (SELECT *,
 (select array_to_string(array_agg(\\\"tag\\\"),
',
')\\n\\t\\t\\t\\tfrom \\\"oc_bookmarks_tags\\\" where \\\"bookmark_id\\\" = \\\"b2\\\".\\\"id\\\") as \\\"tags\\\"\\n\\t\\t\\t\\tFROM \\\"oc_bookmarks\\\" \\\"b2\\\"\\n\\t\\t\\t\\tWHERE \\\"user_id\\\" = ? ) as \\\"b\\\" WHERE true  AND lower(url || title || description || ifnull(tags,
'') ) like ?  ORDER BY id DESC  LIMIT 10 OFFSET 0' with params [\\\"pega\\\",
 \\\"%django%\\\"]:\\n\\nSQLSTATE[42883]: Undefined function: 7 ERROR:  function ifnull(text,
 unknown) does not exist\\nLINE 4: ...RE true  AND lower(url || title || description || ifnull(tag...\\n                                                             ^\\nHINT:  No function matches the given name and argument types. You might need to add explicit type casts.\",
\"Code\":0,
\"Trace\":\"#0 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/3rdparty\\\/doctrine\\\/dbal\\\/lib\\\/Doctrine\\\/DBAL\\\/DBALException.php(116): Doctrine\\\\DBAL\\\\Driver\\\\AbstractPostgreSQLDriver->convertException('An exception oc...',
 Object(Doctrine\\\\DBAL\\\\Driver\\\\PDOException))\\n#1 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/3rdparty\\\/doctrine\\\/dbal\\\/lib\\\/Doctrine\\\/DBAL\\\/Statement.php(172): Doctrine\\\\DBAL\\\\DBALException::driverExceptionDuringQuery(Object(Doctrine\\\\DBAL\\\\Driver\\\\PDOPgSql\\\\Driver),
 Object(Doctrine\\\\DBAL\\\\Driver\\\\PDOException),
 'SELECT * FROM (...',
 Array)\\n#2 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/lib\\\/private\\\/db\\\/statementwrapper.php(63): Doctrine\\\\DBAL\\\\Statement->execute(Array)\\n#3 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/apps\\\/bookmarks\\\/controller\\\/lib\\\/bookmarks.php(177): OC_DB_StatementWrapper->execute(Array)\\n#4 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/apps\\\/bookmarks\\\/controller\\\/lib\\\/search.php(42): OCA\\\\Bookmarks\\\\Controller\\\\Lib\\\\Bookmarks::findBookmarks('pega',
 Object(OC\\\\AppFramework\\\\Db\\\\Db),
 0,
 'id',
 'django',
 false)\\n#5 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/lib\\\/private\\\/search.php(66): OCA\\\\Bookmarks\\\\Controller\\\\Lib\\\\Search->search('django')\\n#6 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/core\\\/search\\\/ajax\\\/search.php(53): OC\\\\Search->searchPaged('django',
 Array,
 1,
 30)\\n#7 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/lib\\\/private\\\/route\\\/route.php(135) : runtime-created function(1): require_once('\\\/var\\\/www\\\/mydomain...')\\n#8 [internal function]: __lambda_func(Array)\\n#9 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/lib\\\/private\\\/route\\\/router.php(250): call_user_func('\\\\x00lambda_166453',
 Array)\\n#10 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/lib\\\/base.php(763): OC\\\\Route\\\\Router->match('\\\/core\\\/search')\\n#11 \\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/index.php(36): OC::handleRequest()\\n#12 {main}\",
\"File\":\"\\\/var\\\/www\\\/mydomain.com\\\/www\\\/htdocs\\\/3rdparty\\\/doctrine\\\/dbal\\\/lib\\\/Doctrine\\\/DBAL\\\/Driver\\\/AbstractPostgreSQLDriver.php\",
\"Line\":91}",
"level":4,
"time":"2015-03-03T07:27:25+00:00"}
```

While this happens, my bookmarks list does not modify (as I never tried to search in first place) and a spinning icon appears down below saying "Searching other places..."

Seeing the logs seems like something I almost I could fix, but I don't quite understand what does this mean:

```
Undefined function: 7 ERROR:  function ifnull(text,
 unknown) does not exist\\nLINE 4: ...RE true  AND lower(url || title || description || ifnull(tag...\\n                                                             ^\\nHINT:  No function matches the given name and argument types. You might need to add explicit type casts.\",
\"Code\":0,
\"Trace\":\"#0 
```

Thanks in advance.

### LINKED ISSUE #157: Bulk editing
Hi
It is very tedious to delete bmarks one by one. It would be useful if it was possible to select multiple bmarks and delete them.

### LINKED ISSUE #168: How do I unfilter my bookmarks?
Once I click on a tag on the left, my bookmarks are filtered to just that tag. How do I get back to the original list of tags without reloading the page?

### LINKED ISSUE #255: Issue importing Bookmarks
Hi all,
when I enabled the bookmarks app and tried to import bookmarks nothing worked. I figured out a workaround, but want you to know of this issue.

When inspecting the import button you get
`<input type="file" id="bm_import" name="bm_import" size="5">`
and further down the road 
`<button type="button" name="bm_import_btn" id="bm_import_submit">Importieren</button>`

What I needed to do in order for the button to work, was changing 
`$('#bm_import').change(attachSettingEvent);`
in bookmarks.js to use the name of the button:
`$('#bm_import_btn').change(attachSettingEvent);`

Then the import dialogue opens as expected.

I'm pretty unsure about everything js and html, so sorry for not creating a fix (if there is an issue :smile: ). Hope this helps.

Cheers, DeastinY
### Steps to reproduce
1. Open Bookmarks (non jet imported)
2. Open Sidebar
3. Click Import
### Expected behaviour

Import-dialogue should open
### Actual behaviour

Nothing happens
### Server configuration

**Operating system**:  
Debian 3.16.7
**ownCloud version:** (see ownCloud admin page)
9.0.1 (stable)
**Bookmarks version:** (see ownCloud apps page)
0.7
**Updated from an older ownCloud or fresh install:**
Updated
### Client configuration

**Browser:**
Current Firefox/Chrome/Chromium
**Operating system:**
Current Ubuntu

### LINKED ISSUE #257: URL metadata extraction causes an exception and results in a 500 response
### Steps to reproduce
1. Trying to book the following urls causes an error:
2. https://software.intel.com/en-us/android/articles/unity-optimization-guide-for-x86-android-part-1 (certificate issue)
3. http://www.nytimes.com/2016/05/29/opinion/sunday/why-you-will-marry-the-wrong-person.html (Will not follow more than 6 redirects)

The issue seems to be cause by this line:
https://github.com/owncloud/bookmarks/blob/master/controller/rest/bookmarkcontroller.php#L75

While I appreciate getting some data from the url, if there is an issue with the `getURLMetadata` call, the app should skip this part and still save the bookmark.

I am happy to submit a PR if this solution is acceptable.
### Server configuration

**Operating system**: Debian Jessie

**Web server:** Nginx

**Database:** MySQL

**PHP version:** 5.6

**ownCloud version:** 9.0.2

**Bookmarks version:** 0.7

**Updated from an older ownCloud or fresh install:** fresh install with data migration from latest 8.2

**Where did you install ownCloud from:** apt package manager

**Signing status (ownCloud 9.0 and above):** Seems to be OK

### LINKED ISSUE #286: Workaround patch for no_tag filtering
Hi! Several people reported issues with annoying scrolling when trying to tag a lot of untagged bookmarks.
#41 #12

I've done a quick patch for no_tag filter to make bookmarks app a little more usable. (apply to current master)
Usage: write "_notag" in the tags filter search bar.

```
diff --git a/controller/lib/bookmarks.php b/controller/lib/bookmarks.php
index e0fe466..8baf6c4 100644
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -187,6 +187,20 @@ class Bookmarks {
        return $bookmarks;
    }

+   public static function findUntaggedBookmarks($userId, IDb $db, $offset, $limit = 10) {
+       $sql = "SELECT * FROM `*PREFIX*bookmarks` where `id` not in (select `bookmark_id` from `*PREFIX*bookmarks_tags`) AND `user_id` = ?";
+
+       if ($limit == -1 || $limit === false) {
+           $limit = null;
+           $offset = null;
+       }
+
+       $query = $db->prepareQuery($sql, $limit, $offset);
+       $results = $query->execute(array($userId))->fetchAll();
+
+       return $results;
+   }
+
    private static function findBookmarksBuildFilter(&$sql, &$params, $filters, $filterTagOnly, $tagFilterConjunction, $CONFIG_DBTYPE) {
        $tagOrSearch = false;
        $connectWord = 'AND';
index cd6597b..a9d53a5 100644
@@ -66,16 +66,21 @@ class BookmarkController extends ApiController {
            $qtags = Bookmarks::findTags($this->userId, $this->db, $tags);
            return new JSONResponse(array('data' => $qtags, 'status' => 'success'));
        } else { // type == bookmark
-           $filterTag = Bookmarks::analyzeTagRequest($tag);
-
            $offset = $page * 10;

-           if ($sort == 'bookmarks_sorting_clicks') {
-               $sqlSortColumn = 'clickcount';
-           } else {
-               $sqlSortColumn = 'lastmodified';
+           if ($tag == "_notag") {
+               $bookmarks = Bookmarks::findUntaggedBookmarks($this->userId, $this->db, $offset);
+           }
+           else {
+               $filterTag = Bookmarks::analyzeTagRequest($tag);
+
+               if ($sort == 'bookmarks_sorting_clicks') {
+                   $sqlSortColumn = 'clickcount';
+               } else {
+                   $sqlSortColumn = 'lastmodified';
+               }
+               $bookmarks = Bookmarks::findBookmarks($this->userId, $this->db, $offset, $sqlSortColumn, $filterTag, true);
            }
-           $bookmarks = Bookmarks::findBookmarks($this->userId, $this->db, $offset, $sqlSortColumn, $filterTag, true);
            return new JSONResponse(array('data' => $bookmarks, 'status' => 'success'));
        }
    }
```

Hope somebody found this patch useful.

### LINKED ISSUE #292: Show bookmark description and/or icon in overview
### Steps to reproduce
1. Go to bookmarks app
2. See the list of bookmarks
### Expected behaviour

I exepct some way to see the description (and/or the favicon of the link).
Currently the description is only visible when editing or creating a bookmark.
At least it should be possible to "Show descriptions" via an setting/option, if this should be not default.
### Actual behaviour

Just a List with link, tags and date. No description. 
Because of the limited space for the url it's not easy (depending on site's link title sometimes impossible) to identify whats behind this link.

unbenannt
### Server configuration

**Operating system**: Debian 8

**Web server:** Apache

**Database:** MySql

**PHP version:** 5.6

**ownCloud version:** 9.1.0.16 / NextCloud 10.0

**Bookmarks version:**  0.9.0

### LINKED ISSUE #297: Search Bookmarks Capability
This is a requested feature / enhancement  (and my first every request on git)
# **Steps to reproduce:**
1. Go to bookmarks app
2. See hundreds, if not thousands of bookmarks and an overwhelming number of tags
# **Expected Behavior:**

A basic search field in the left sidebar, between the Add Address button and the Filter by tag box.  For those of us with thousands of bookmarks and nearly a hundred tags, having a basic search capability would be an enormously valuable feature.  This should also be very straight forward to implement.
# **Actual behaviour:**

No ability to search bookmarks.  One must first find a relevant tag, then another, and another, which is not easy because they are not listed in alphabetical order (which would make that process easier).  It can take substantial time to find a bookmark because there is no search ability, and tags are not in any proper order on the sidebar.

_To the developers what work on these Apps and the Nextcloud platform, Thank you for your amazing work and dedication._

### LINKED ISSUE #423: Import of large file fails silently
### Steps to reproduce
1. Export larger (Firefox) bookmarks HTML file (~5000 lines)
2. In Chrome, import into Bookmarks

### Expected behaviour
Bookmarks should be imported or an error message should appear.

### Actual behaviour
Nothing happens in the UI.

The import actually fails with this error in the browser console:
```
Failed to load resource: the server responded with a status of 413 (Request Entity Too Large)
```

Bookmarks 0.10.1
Chrome 63 on MacOS

### LINKED ISSUE #430: Can you optimize page to view in sidebar?
I'm talking about sidebar in new firefox quantum. On my screen it takes about 370 pixels, so it all comes down to optimize to mobile view.
image

### LINKED ISSUE #437: Pressing enter while editing bookmark's title cancels edition
### Steps to reproduce
1. Click on the button to edit a bookmark next to it
2. Edit the title
3. Press Enter

### Expected behaviour
Changes are saved

### Actual behaviour
Changes are lost without even being prompted

### Server configuration

**Nextcloud version:** 13.0.0 RC 4

**Bookmarks version:** 0.10.1

marcelklehr:
screenshot from 2018-01-28 14-57-16

- [x] adding bookmarks (fixes #257)
- [x] bookmark search (fixes #131, fixes #297)
- [x] bulk editing (fixes #157)
   - [x] bulk deletion
   - [x] bulk tagging (fixes #12; fixes #286)
- [x] Import & export
   - [x] Import: use fetch() API if available (fixes #423, fixes #80)
   - [x] Use new upload ui (fixes #255 )
- [x] Responsive design (fixes #430)
- [x] editing bookmarks (fixes #437)
- [x] editing tags
- [x] website images (fixes #292)
  - [x] grid view (fixes #55)
- [x] translatability
- [ ] accessibility
- [x] advanced tag filter (fixes #168)
   - [ ] 'Untagged' menu item (do we really need this when we have bulk editing and no more reloads?)
- [ ] ~~sharing (#5)~~

Installation:

```
git clone #-this-repo-#
git checkout feature/new-ui
composer install
npm install
npm run build
```

*old pr: #402*


splitt3r:
Nice :) the address field would look pretty awesome if it would fit the navigation style like:

add

And the settings without the button look like:

settings

marcelklehr:
Thanks :) Good point on the settings :+1: 

re the "new bookmark" form: I originally intended to make it borderless, but wasn't sure whether this would be confusing to people who maybe wouldn't be able to tell that it's a text input. Happy to hear feedback on this, as well as anything else :)

splitt3r:
It´s not a borderless input it´s an anchor that works like https://docs.nextcloud.com/server/12/developer_manual/design/navigation.html#edit-entry :) More and more apps are adopting the new style guidelines so the user will get used to it and love it ❤️ i think :D

marcelklehr:
Ah, I see, that's nice indeed. Done :)

marcelklehr:
Quick question: I have now added a special tags view that allow the user to edit, rename and filter by tags. Do you people think, the "favorite tags" should link there directly, with the clicked tag already selected, or should those go into a simple view without all the tag filtering and mangement goodies?

(cc @jancborchardt @Espina2 )

jancborchardt:
@marcelklehr the tags should _all_ be listed in the left navigation, sorted by most used. No need for an additional tagging view.

The tag list works like the navigation usually: Clicking on a tag shows all bookmarks with that tag. Clicking another tag switches to that other tag. No combining for now as that led to confusion in the other version.

Let's get this design change out as soon as possible, and let's keep things as simple as possible. :)

marcelklehr:
Mh. I think it's likely that people will have *a lot* of tags, so I'm not sure displaying them all in the sidebar is a good idea. (Which is why we opted for *favorite* tags only, as far as I can remember.) Also, I must agree with arthur, in that the whole point of having tags, to me, seems to be that you're able to combine them. Otherwise they're even less useful than folders. After reading #168, I think the problems with the old implementation of filtering were basically:
 - modes in the sidebar
 - tags disappear if you change the filter ("drilldown feature")
 - issues with resetting the app

The implementation I have right now, does neither of these things. Here's a screenshot, for example:
screenshot from 2018-01-22 14-01-51

jancborchardt:
Still, why not have all tags in the sidebar? Normal users have around 5-10 tags, and those easily fit in the navigation. Even 20 or more would work nicely with the sorting by number of bookmarks tagged with it.

Combining could even be possible there easily, with an entry in the 3-dot-menu of a tag for it, along with "Rename" and "Delete".

Remember that we are designing Nextcloud for regular people. Our competition or benchmark is stuff like Pocket, or bookmarking in the browser. 

The main point of having tags is to bring some order to your collection. Combining is very advanced, and definitely not the whole point. :)

pixelipo:
Regarding tags, I agree with @jancborchardt - I think they should all be listed in a single sidebar.

Another thing - "Tags" in the left sidebar should not be selectable. It should simply be a title for the subsequent list of tags.

And lastly, unrelated to tags, why are bookmarks square boxes? I see no good reason to not have them in a normal vertical list, just like a file list. It seems like all the content would even fit in a single line. Eyes work much faster on a sorted list than a grid, especially on something so non-visual, like this.

Grid would only make sense if the plan is to fetch previews of the websites.

marcelklehr:
Thanks for your feedback, folks! :)

The 'tags' menu item is a relic of the specialized tags view, I introduced. I agree that it would be best to have the tags management all in the sidebar instead of a new view, though.

Another quick question on this: What's your opinion on having all selected tags stick to the top of the list? Might be confusing if the locations change.

The idea was to have previews yes. Then it turned out it's quite hard to realize, so we settled on touch icons and that sort of stuff. I'm not sure how best to display those, though, especially since not every site has them. I agree that without visuals it'd be better to have a list. After the tags management I'm going to work on the visuals and see how that pans out. If it doesn't, going back to a list view seems like the best option indeed.

marcelklehr:
screenshot from 2018-01-23 23-20-24

This is how it looks now. :)

jancborchardt:
> Another quick question on this: What's your opinion on having all selected tags stick to the top of the list? Might be confusing if the locations change.

@marcelklehr yeah, as you already say that would be confusing. :) The tags should be sorted by number of bookmarks tagged with it, and so left of the 3-dot-menu there should also be a counter displaying the number of bookmarks.

@pixelipo yep, as @marcelklehr mentioned the plan is to use icons (as preview fetching is hard). A list would be pretty bland.

jancborchardt:
@marcelklehr I will look into style detail improvements. Could you check to add the counter to the tags, and also maybe a first quick implementation of the favicon stuff?
Maybe @jakobsack @christophwurst have some pointers here as they did favicon fetching for the Mail app. :)

marcelklehr:
I've found a nice drupal module that handles link preview pretty well (fetches meta tags and the largest images in the html as a fallback): https://www.drupal.org/project/external_link_preview
But of course any experience from the mail app team is helpful! :)
I'll look into adding counters, too.

ChristophWurst:
> Maybe @jakobsack @ChristophWurst have some pointers here as they did favicon fetching for the Mail app. :)

https://github.com/nextcloud/mail/pull/469/files 😉

marcelklehr:
Is there a way to prevent a composer dependency of an app (in this case guzzle) to conflict with an already installed version of that library?

marcelklehr:
Thanks @ChristophWurst that was very helpful! I didn't know about ICache :)

screenshot from 2018-01-28 14-57-16

One problem remains: The dependency I used to fetch the metadata depends on a different version of guzzle which conflicts with the one nextcloud has installed by default... :/

ChristophWurst:
> Is there a way to prevent a composer dependency of an app (in this case guzzle) to conflict with an already installed version of that library?

No, there is no way to prevent dependency hell in php. I suggest to use the guzzle from server. What do you need it for? Is the `IClientFactory` interface sufficient?

marcelklehr:
A library I wanted to use depends on a different version of guzzle, so I forked and backported the library to use guzzle v5.3 :/

pixelipo:
I'm not sure it's such a good idea to have the entire frontend in a single file: `js/bookmarks.js`

It would be nice to have it split up into more files, perhaps moving views outside as well...

marcelklehr:
@pixelipo Dead right. That was merely for development, because I didn't want to deal with build steps and all that stuff :D

templates/main.php
```diff
@@ -1,8 +1,12 @@
 <?php
+script('bookmarks', '3rdparty/backbone');
```
ChristophWurst: FYI this will potentially cause problems (either in your app, server components or other apps that are loaded on the same page) as you're overwriting a global js lib, assuming you're installing a newer version of Backbone than the one loaded by the server.

If you go with a build tool like webpack, you can bundle your libraries into a single js that doesn't conflict with other scripts.

jancborchardt:
@marcelklehr nice progress! :) Looks much better already now! :tada:

templates/main.php
marcelklehr: Very true :) Tackled modularization and bundling just now ;)

marcelklehr:
I think sharing bookmarks is something for a different pr. I've also disabled favorites for now, since there's no API for this, yet (and the objective seems to be to get this out ASAP ;) ).

Other than that, I think it's ready for a first review :)

One remark: I didn't even try to comply with the obligatory use of semicolons in js because I wouldn't have been very successful :see_no_evil: , so I've mostly stuck to my usual coding style and am going to convert the code, once it passes your reviews :D

marcelklehr:
(I've added instructions above on how to build the code :D)

jancborchardt:
Cc @nextcloud/javascript and @nextcloud/designers for review :)

js/templates/Settings.html
```diff
@@ -0,0 +1 @@
+<div id="app-settings-header"><button class="settings-button">Settings</a></div><div id="app-settings-content"><form class="import-form" action="bookmark/import" method="post" target="upload_iframe" enctype="multipart/form-data" encoding="multipart/form-data"><input type="file" class="import" name="bm_import" size="5" /><input type="hidden" name="requesttoken" value="'+oc_requesttoken+'" /><button class="import-facade"><span class="icon-upload"></span> Import</button></form><iframe class="upload" name="upload_iframe" id="upload_iframe"></iframe><button class="export"><span class="icon-download"></span> Export</button><div class="import-status"></div></div>
```
pixelipo: prettify maybe? Whole template in a single line is not very dev-friendly

css/bookmarks.css
```diff
@@ -1,435 +1,429 @@
...
-
-#add_form {
-    padding-bottom: 2em;
-}
-
-#add_form_loading {
-    visibility: hidden;
-    float: right;
-    margin: 8px;
+#content > div {
```
pixelipo: Both CSS files should be prettified to meet NC coding standards.

js/templates/Settings.html
marcelklehr: I agree!

marcelklehr:
@pixelipo done.

css/select2.css
```diff
@@ -0,0 +1 @@
+
```
pixelipo: I guess you can safely delete this file now?

templates/main.php
```diff
@@ -1,12 +1,8 @@
...
-script('bookmarks', 'settings');
-script('bookmarks', 'bookmarks');
+script('bookmarks', 'dist/bundle');
 style('bookmarks', 'bookmarks');
 
-script('bookmarks', '3rdparty/tag-it');
-script('bookmarks', '3rdparty/js_tpl');
-style('bookmarks', '3rdparty/jquery.tagit');
-
+style('bookmarks', 'select2');
```
pixelipo: `'select2'` is empty, so not needed, right?

pixelipo:
Great work, @marcelklehr - code looks really clean now! I will test it on a real NC installation on Monday (can't make it earlier, sorry). Maybe others can...

css/select2.css
marcelklehr: uh? that was not my intention, actually :D

marcelklehr:
Also whoever merges this, should probably squash it, no?

pixelipo:
@marcelklehr usually it's the author that is responsible for squashing in NC

js/templates/BookmarkDetail_default.html
```diff
@@ -0,0 +1,9 @@
+<div class="close icon-close" title="<%- t('bookmarks', 'Close') %>"></div>
```
pixelipo: Why is this file not named using PascalCase, like all the other html files?

js/templates/BookmarkDetail_editing.html
```diff
@@ -0,0 +1,14 @@
+<div class="close icon-close" title="<%- t('bookmarks', 'Close') %>"></div>
```
pixelipo: same as above

js/templates/BookmarkDetail_default.html
marcelklehr: Ah. The idea was to signify that the two files are different modes of the same view. I'm not particular about that, though. If you like, I can just change that to PascalCase.

pixelipo: @marcelklehr I have no strong opinion on that, if `BookmarkDetail_default` is ok for everyone, then leave it be :)

9662:
@jancborchardt 

> The main point of having tags is to bring some order to your collection. Combining is very advanced, and definitely not the whole point. 

FWIW, see this [example story](https://github.com/nextcloud/bookmarks/issues/5#issuecomment-366564046). I would reasonably expect to be able to filter by something like `project-b` + `materials` for example.

9662:
@marcelklehr 

> The idea was to have previews yes. Then it turned out it's quite hard to realize, so we settled on touch icons and that sort of stuff. I'm not sure how best to display those, though, especially since not every site has them.

You could try:
 * Scan bookmark target for JSON-LD, OpenGraph metadata to see if there is an [image](http://schema.org/image) or [logo](http://schema.org/logo) property and fetch that.
 * Scan bookmark target for a `<link rel="icon" …>` element and use that.
 * Blindly go for `/favicon.ico` (does anyone still use that?).
 * Show a generic image.

Personally, I'm just fine with having a vanilla list with one bookmark per row though.

marcelklehr:
@9662 Thanks for your feedback.We're indeed using a library to extract the metadata and image from the URLs now :) In the current version of this pr, it's possible to select multiple tags to filter by, as can be seen in https://github.com/nextcloud/bookmarks/pull/424#issuecomment-359953112 ;)

jancborchardt:
@marcelklehr since it's such a big rewrite it's not necessary to squash. It will help later in case there are some bugs and make it easier to find what caused it. :)

@nextcloud/javascript @blizzz so is this good to go? :)

marcelklehr:
The only test that is still failing is this one, but it doesn't really make sense to me in the first place: https://github.com/nextcloud/bookmarks/blob/dc86ef0bc21ce15388055e4abbc17a05f12f1b39/tests/bookmarkcontroller_test.php#L200 (I know I was one of the last people that touched this part, but I didn't check whether it made sense back then. Shame on me :D)

(Why doesn't it make sense? `%` is an escape character. Registering a url with a non-escaped escape characer and then expecting it to be recognized when that escape character *is* escaped, shouldn't work as those are not the same URLs, IMHO)

sunjam:
Well it be possible to view the bookmarks as a compact list ala _Show all Bookmarks_ in Firefox?
    
80656843-0361-401b-a286-29071100be88

jancborchardt:
@sunjam no. We are moving forward with this new design, and maintaining two different views is too much. Besides, the compact list is not well-designed:
- takes a lot of width, so not really compact at all.
- title and link are very far away from each other.
- row height is very low so it's not easily clickable.

jancborchardt:
@marcelklehr how about we merge this and fix things going forward? It's not a release anyway. But if we keep counting this up to 100 commits, it will never get in. ;)

marcelklehr:
@jancborchardt heh, personally I'm happy to merge it, however, I'm not sure anybody besides me has tested the code yet and also, usually @blizzz gets to leave his input about stuff I submit. :)

pixelipo:
1. @marcelklehr why is the checkbox inside the 3-dot menu? It's already available on the item itself - no need to duplicate:
image

2. Icon should be white:
image

3. `textarea` should have `resize:vertical` property (disable horizontal resizing):
image

4. Is there any need for this view in the sidebar? I think that the "Edit" view should be the only view. Ideally, editing would be "inline" and autosave should also exist, but I understand if that is out of scope for this PR.
image

5. Settings menu looks, hmmm, strange:
image

6. Question for @jancborchardt @skjnldsv : button in sidebar - what is the design guideline for this? It's not defined in the [guidelines](https://docs.nextcloud.com/server/13/developer_manual/design/navigation.html) yet, and we are currently seeing a lot of different solutions there...
image
image

7. Few issues regarding the multiselect form (great solution, by the way!):
image
- lightbue background color does not really fit the design - epsecially when you look at the white arrow pointing to the Bookmarks app icon
- select area is not using default styling (for example, border becomes black on focus)
- "x" button on the right is too small; actually I think it should be removed completely, because it's a very fringe use-case (if you want to rapidly remove a ton tags) and very destructive in nature) CC @jancborchardt 
- "x" button on the tags is too small (see Deck app for comparison)
- "Select all.." icon is too close to the button text
- clicking on a bookmark while the multiselect is visible causes this:
image
- it looks really messy in mobile view:
image

8. This menu looks weird. 
image
In fact, I think that checkbox should be moved out of the 3-dot and in the usual position of the icon (left of text), but we haven't discussed this use-case in @nextcloud/designers Something like this?
image

***

I think some of the issues I found are serious enough that if the app is deployed as-is, I would call it "alpha"

pixelipo:
off topic - I'm unable to test images, because they are not loaded for some reason. @marcelklehr do you have any idea why? Links definitely look weird:
image

skjnldsv:
> Question for @jancborchardt @skjnldsv : button in sidebar - what is the design guideline for this? It's not defined in the guidelines yet, and we are currently seeing a lot of different solutions there...

Guidelines: https://github.com/nextcloud/documentation/pull/684
But it's waiting to be merged here: https://github.com/nextcloud/server/pull/7600

pixelipo:
@skjnldsv in the nextcloud/server#7600 "Add group" button is different - is that by design or planned for upgrade?

skjnldsv:
@pixelipo it was by design. Last time it was discussed, the outcome was that only one button is supposed to be displayed. :)

Not sure if we want to switch to a secondary button for new groups? :thinking: 
It's not THAT bad tough :)
capture d ecran_2018-02-27_13-14-56

MariusBluem:
I thought we could kill the Add group button at all ... You can create groups by assigning a non-existing group name to a user already ;) 2 action buttons in the sidebar are to noisy I think.

skjnldsv:
I agree @MariusBluem :)

marcelklehr:
Thanks for taking the time, Marin :)

1. I think the objective was to allow access to selection to users on touch devices, who don't have hover, as the checkbox on the card is display on hover only.
2. fixed.
3. fixed.
4. The point of the sidebar, besides editing is the ability to view the description, which might turn into a read-it-later-style functionality some day :) I agree that it's not ideal, though. Autosave is not hard to do.
5. fixed :D 
6. I'm not particular about this :) Originally I had a button, then someone suggested the solution as it is now.
7. I agree on all your points here. I'll change the background and fix the styling and the "x"es. That sidebar / bulk editing clash looks bad, indeed. Do you think it would make sense to wire the click on a bookmark card to make it selected without opening the sidebar?
8. Fixed the padding, although your suggestion also looks nice.

marcelklehr:
Re: images: The URLs look ok to me. They are hitting an API endpoint that returns the image for a bookmark id. If you're getting 404s on those URLs, that means there was no image found for that site. Maybe try something like a youtube video -- that works for me at least.

sunjam:
Hi, my instance is currently offline due to disk corruption, so I cannot test the latest build, but I was able to make your request for testing this pull request the top pinned issue on [help.nextcloud.com](https://help.nextcloud.com), so you may soon receive a number of new testers.  Thanks for your great work on the app!

pixelipo:
Great work, @marcelklehr !

1. That makes sense - we shouldn't hide actions behind `:hover`. So why not just always show
4. Updated sidebar looks really nice and modern. I do have a couple of issues with it:
  * Delete button doesn't work for me (the one in the 3-dot menu works, though); Maybe it's not even needed? Do we need "Delete" in two different places?
  * "description" doesn't have a visible "edit" visual cue. My suggestion:
```
.description:empty::before {
    content: "Add a description";
    opacity: .3;
}
.description {
    cursor: text;
}
```
This is not translatable, though. String should be injected otherwise.
6. I'm not particular about that button, either. You can keep it for now, or update it, as you wish
7. If you mean "wire it to select only when multiselect popover is visible", then yes, that might be a good idea. It will cause Description not to be accessible during selecting, but that's a fair tradeoff, I think.
8. I would vote for moving the checkbox out of the 3-dot menu and to the left of the tag name. It will half the number of clicks and cursor trips, with no real downside...

Also: Delete button doesn't have a confirmation dialog nor an "Undo" option. It would be worth considering adding one of those (second one being better). Maybe in a future PR?
Autosave feature should have a visual cue that data is saved (check the Deck app for a similar functionality)

guddl:
If you have many TAGs -  there should be a scrollbar at the sidebar. On my big screen not all TAGs are shown :grin:

marcelklehr:
@guddl Thanks for the feedback! It's fixed now. :)

marcelklehr:
@pixelipo Did you get the bookmark images to work? (I also added some URL-based colors for the ones without images, btw)

marcelklehr:
Thanks for your feedback folks! :)

jancborchardt:
@marcelklehr yaaaayyyy, finally! :) Congrats!

jospoortvliet:
Ok, we should do something with this, blog with some screenshots for example!

marcelklehr:
@jospoortvliet Good idea! :) It's not released yet, though, so I'm not sure if it'll be that valuable for the audience. I think a better time for a blog article would probably be the next release, when people can actually install and try the new experience. :)

jospoortvliet:
ok, get me some basic info on what changed a few days before the next release! You can write a nice blog if you like but you don't have to - a bullet list of changes is enough, we can write something nice! In this case, links to PR's is actually good enough - and this issue seems an overview so perhaps this will suffice, if it has everything that's new.

Then a few screenshots and we're done :rocket:

bunducafe:
Sorry for a kind of off topic question: I activated the bookmark plugin via the nextcloud admin interface. But I cannot see any of the new UI features.
What I get is on the left side just a field for adding an address and the search box for the tags. No bulk deletion, no "all bookmarks", nothing is shown... 

how can I solve this?

marcelklehr:
@bunducafe  I'm sorry to tell you that the latest features have not been released yet, but merely staged for inclusion in the next release. The README does indeed not correctly reflect this. Currently, you can only get the new features by manually installing the app (as described in the README).
Don't be afraid to open a new issue for any questions you might have :)

@jospoortvliet Thanks! I will get back to you! :)

e-alfred:
@marcelklehr A dead link check would be a great feature I think. Maybe links that throw an error are marked by a tag for example. It could be either done periodically or just with a button to start the check manually.

Issue #425, opened 2018-01-12
CI fails with nc master


marcelklehr:
Reason: master apparently dropped support for php 5.x :(

blizzz:
yes, Nextcloud 14 will need PHP >=7

PR #426, opened 2018-01-14
Do not use deprecated methods anymore

MorrisJobke:
See https://github.com/nextcloud/server/pull/7857

PR #427, opened 2018-01-15
include nc 13 in travis test and adjust cases accordingly

blizzz:
to solve #425 

let's see whether i caught everything

### DIFF
--- a/templates/js_tpl.php
+++ b/templates/js_tpl.php
@@ -2,7 +2,7 @@
     <div class="bookmark_single" data-id="<&= id &>">
         <p class="bookmark_actions">
             <span class="bookmark_delete">
-                <img class="svg" src="<?php print_unescaped(OCP\image_path("", "actions/delete.svg")); ?>"
+                <img class="svg" src="<?php print_unescaped(image_path("", "actions/delete.svg")); ?>"
                      title="<?php p($l->t('Delete')); ?>">
             </span>&nbsp;
         </p>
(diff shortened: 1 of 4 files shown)


templates/js_tpl.php
```diff
@@ -11,7 +11,7 @@
                 <&= escapeHTML(title == '' ? encodeURI(url) : title ) &>
             </a>
             <span class="bookmark_edit bookmark_edit_btn">
-                <img class="svg" src="<?php print_unescaped(OCP\image_path("", "actions/rename.svg")); ?>" title="<?php p($l->t('Edit')); ?>">
```
marcelklehr: (see #426 ;))

blizzz:
fixed with recent commit :)

templates/js_tpl.php
```diff
@@ -51,11 +51,11 @@
     <li><a href="" class="tag"><&= escapeHTML(tag) &></a>
         <div class="tags_actions">
             <span class="tag_delete">
-                <img class="svg" src="<?php print_unescaped(OCP\image_path("", "actions/delete.svg")); ?>"
+                <img class="svg" src="<?php print_unescaped(image_path("", "actions/delete.svg")); ?>"
                      title="<?php p($l->t('Delete')); ?>">
             </span>
             <span class="tag_edit">
-                <img class="svg" src="<?php print_unescaped(OCP\image_path("", "actions/rename.svg")); ?>"
```
marcelklehr: Is this a good idea? I have no clue how the templating works, though. From this I assume, it's a separate scope, which would give this a *lgtm* verdict.

marcelklehr (approved):
lgtm :ship:

templates/js_tpl.php
blizzz: yeah, he methods are defined in ``lib/private/legacy/template/functions.php`` and also called by the Template class. These short forms are also used in server and shipped apps.

PR #452, opened 2018-03-15
possibility of sharing

hercut:
It would be interesting to give the possibility to share folders with other nextcloud users.

This makes it possible to have favorites managed by several people of a team.


blizzz:
there's already an issue…

PR #469, opened 2018-04-11
Update bookmarkcontroller.php to allow bookmarks import using curl utility (api)

milosz:
It is possible to export using curl (api), but import is not possible due to csrf (api).


marcelklehr:
Thanks for the fix!

Once the tests pass this is good to go :)

PR #473, opened 2018-04-18
Fix database row size

### LINKED ISSUE #472: Build error on latest git release of Bookmarks via Node.js 9.x
Built Bookmarks app latest git release on *Debian 9 Stretch* using latest Node.js 9.x, but when I attempt to enable it in *Nextcloud Stable 13.0.1* it fails with the following error:

An exception occurred while executing 'ALTER TABLE oc_bookmarks ADD `image` VARCHAR(4096) DEFAULT NULL, ADD `favicon` VARCHAR(4096) DEFAULT NULL': SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs

marcelklehr:
shrink image and favicon columns as they are only URLs anyway

should fix #472

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -78,14 +78,14 @@
 				<type>text</type>
 				<default></default>
 				<notnull>false</notnull>
-				<length>4096</length>
+				<length>512</length>
 			</field>
 			<field>
 				<name>favicon</name>
 				<type>text</type>
 				<default></default>
 				<notnull>false</notnull>
-				<length>4096</length>
+				<length>512</length>
 			</field>
 		</declaration>
 	</table>


marcelklehr:
@sunjam this should fix your issues with database row size. Could you test to confirm?

sunjam:
Deleted bookmarks, re-installed and getting exact same error.  Just confirming I installed this branch correctly...

`git checkout -b fix/database-row-length`
`composer install`
`npm install`
`npm run build`

marcelklehr:
@blizzz The expected value for the image/favicon URLs is not that large. The actual images are cached elsewhere. Perhaps we can convert the description into a TEXT/BLOB to get the row size down? Is it possible to migrate existing instances in this case?

marcelklehr:
@sunjam `git checkout -b fix/database-row-length` creates a new branch. It depends on your setup, whether git will realize that there is an upstream branch with the same name. See https://stackoverflow.com/questions/1783405/how-do-i-check-out-a-remote-git-branch#1783426

sunjam:
@marcelklehr That is the guide I followed previously as well.  I’m trying to follow instructions from the bookmarks github page + the linked guide with no success.

I successfully installed Node.js 9.x and npm
`Npm install backbone.marionette underscore` after receiving a complaint of missing dependencies.

`git branch -v -a` Does not display your *fix* branch.  I do not know enough about git to resolve this.
Using a best guess I tried `git checkout -b database-row-length fix/database-row-length`
But all I received back is *fatal: not a git repository*

I tried `git checkout -b fix/database-row-length` returned *Switched to a new branch*, then completed composer install, npm install, npm run build.

marcelklehr:
@sunjam This should work:

```
git clone https://github.com/nextcloud/bookmarks.git
cd bookmarks
git fetch origin # fetches the branches of the remote git repository 'origin'
git checkout -b fix-database-row-length origin/fix/database-row-length # Copies the remote branch 'fix/database-row-length' of the remote repo 'origin' to a local branch 'fix-database-row-length'
```
after that you can proceed with installing the javascript dependencies and building the client code with npm as documented in the README.

sunjam:
@marcelklehr 
Bookmarks app now installs and enables, but internal server errors stop it from running.  Bookmarks cannot be added or imported.  There is a problem with npm.  I’ve tried `npm install backbone.marionette backbone underscore`
Also tried global `-g`  installation, but this warning will not go away.  Either warning is for unmet `backbone` or `underscore` peer dependency.

npm WARN backbone.marionette@3.5.1 requires a peer of underscore@~1.8.3 but none is installed. You must install peer dependencies yourself.
npm WARN bookmarks@0.10.1 No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
+ underscore@1.9.0
+ backbone.marionette@3.5.1

marcelklehr:
Mh. Those npm messages are just warnings and they're all non-critical I'd say (it's probably complaining about underscore because it's the wrong version. It should work nontheless).
@sunjam Could you get a hold of the error messages from the nextcloud log that caused those HTTP 500 Internal Server error responses?

blizzz (approved):
Looks good here

PR #474, opened 2018-04-18
Minor UI fixes

marcelklehr:
- remove overflow hidden from bookmarks container
- remove cursor: pointer from bookmark cards as you can only click the title
- add tooltip with tag name on hover
- make detail view scrollable


blizzz (approved):
When opening detail view, the popup stays open.  But that was the case already before.

PR #487, opened 2018-05-13
Fix thumbnail colors

marcelklehr:
- choose from <del>256</del> 500 colors
- loose the alpha channel
- choose color based on domain only

### DIFF
--- a/js/views/BookmarkCard.js
+++ b/js/views/BookmarkCard.js
@@ -8,13 +8,14 @@ import colorPalettes from 'nice-color-palettes';
 const Marionette = Backbone.Marionette;
 const Radio = Backbone.Radio;
 
+// 100 palettes * 5 colors = 500 colors
 const COLORS = colorPalettes.reduce((p1, p2) => p1.concat(p2), []);
 const simpleHash = (str) => {
 	var hash = 0;
 	for (var i = 0; i<str.length; i++){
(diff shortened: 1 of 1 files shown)


js/views/BookmarkCard.js
```diff
@@ -51,7 +52,7 @@ export default Marionette.View.extend({
 		if (this.model.get('image')) {
 			this.$el.css('background-image', 'url(bookmark/'+this.model.get('id')+'/image)');
 		} else {
-			this.$el.css('background-color', COLORS[simpleHash(this.model.get('url')) & 63] + '66');
+			this.$el.css('background-color', COLORS[simpleHash(new URL(this.model.get('url')).host) % 499]);
```
blizzz: where does the 499 come from?

marcelklehr: It's from the [number of colors](https://github.com/nextcloud/bookmarks/pull/487/files#diff-a9872638d77bc8df5f891f028e37c89dR11),probably best to not hardcode it, heh

blizzz: ok, roger. then, i think we do not need to subtract 1? modulo by 500 will still return 499 as max number. Or, is there one color to avoid?

marcelklehr: ah, mind-o :D

blizzz (approved):

PR #491, opened 2018-05-16
wrong version doesn't trigger Doctrine schema updater. Added .gitignore

damko:


damko:
you are right, I'll start using feature branches.

Talking about versions, I wish we had tagging consistent with the package version. It takes me too long to understand the changes between master and package

marcelklehr:
AFAIK, every app release has an associated tag, e.g. https://github.com/nextcloud/bookmarks/tree/v0.11.0 is the latest release

damko:
You are right and I see where my confusion comes from:

screenshot-2018-5-30 nextcloud bookmarks

See the tag list? it doesn't show any 0.10 or 0.11 or 0.9
I think this is due to old tags which started with v8 or v9 therefore the current tagging v0.11 goes very low in the tag list.

I see that it's much better to refer to https://github.com/nextcloud/bookmarks/tags or https://github.com/nextcloud/bookmarks/releases 

Maybe this info should be added in the README?

PR #497, opened 2018-05-25
Release v0.12.0-rc1

marcelklehr:


blizzz:
Can we kill  bookmarks/js/bookmarklet_old.js?

blizzz:
Also when building, we can exclude everything in js, except for dist, correct?

marcelklehr:
PHP 5 support is fixed.

blizzz:
Another thing we can/should do: raise min-Nexcloud version to 12. All earliere series are EOL.

PR #503, opened 2018-06-02
Allow Sorting

chiotisn:
My work, so far, in sorting functionality (#3). 

It works, however, I have problem with running the tests with phpunit (I have post my problem at issue #3)
I will also write a test for findBookmarks function in lib/Bookmarks and I will remove all these "sqlSortColumn" attributes in the controllers that are not needed anymore.

What do you think? Should I fix/improve/add something?

### DIFF
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -209,7 +209,9 @@ class Bookmarks {
 			$qb->selectAlias($qb->createFunction('GROUP_CONCAT(' . $qb->getColumnName('t.tag') . ')'), 'tags');
 		}
 		
-		$sqlSortColumn = $this->config->getUserValue($userid, 'bookmarks', 'sorting', 'lastmodified');
+		if (!in_array($sqlSortColumn, $tableAttributes)) {
+			$sqlSortColumn = 'lastmodified';
+		}
 
 		$qb
 			->from('bookmarks', 'b')
--- a/controller/lib/bookmarks.php
+++ b/controller/lib/bookmarks.php
@@ -208,9 +208,8 @@ class Bookmarks {
         }else{
 			$qb->selectAlias($qb->createFunction('GROUP_CONCAT(' . $qb->getColumnName('t.tag') . ')'), 'tags');
 		}
-		if (!in_array($sqlSortColumn, $tableAttributes)) {
-			$sqlSortColumn = 'lastmodified';
-		}
+		
+		$sqlSortColumn = $this->config->getUserValue($userid, 'bookmarks', 'sorting', 'lastmodified');
 
 		$qb
 			->from('bookmarks', 'b')
@@ -226,7 +225,12 @@ class Bookmarks {
 			$this->findBookmarksBuildFilter($qb, $filters, $filterTagOnly, $tagFilterConjunction);
 		}
 
-		$qb->orderBy($sqlSortColumn, 'DESC');
+		if ($sqlSortColumn == 'title') {
+			$qb->orderBy($qb->createFunction('UPPER(`title`)'), 'ASC');
+		} else {
+			$qb->orderBy($sqlSortColumn, 'DESC');
+		}
+
 		if ($limit != -1 && $limit !== false) {
 			$qb->setMaxResults($limit);
 			if ($offset != null) {
--- a/js/models/Bookmarks.js
+++ b/js/models/Bookmarks.js
@@ -10,6 +10,18 @@ export default Backbone.Collection.extend({
 	parse: function(json) {
 		return json.data;
 	},
+	//override Backbone#comparator
+	comparator: function(m) {
+		if (this.sortby == 'title') {
+			return m.get('title').toLowerCase(); //for case insensitive sorting 
+		} else if (this.sortby == 'added') {
+			return (-1)*m.get('added'); //for descending sorting
+		} else if (this.sortby == 'lastmodified') {
+			return (-1)*m.get('lastmodified');
+		} else if (this.sortby == 'clickcount') {
+			return (-1)*m.get('clickcount');
+		}
+	},
 	initialize: function() {
 		this.loadingState = new Backbone.Model({
 			page: 0,
@@ -33,12 +45,11 @@ export default Backbone.Collection.extend({
 		}
 		const nextPage = this.loadingState.get('page');
 		this.loadingState.set({page: nextPage+1, fetching: true});
-		var sortby = this.sortby;
 		// Show spinner after 1.5s if we're fetching a new query
 		const spinnerTimeout = setTimeout(() => nextPage === 0 && this.reset(), 1500);
 
 		return this.fetch({
-			data: _.extend({}, this.loadingState.get('query'), {page: nextPage, limit: BATCH_SIZE, sortby: sortby}),
+			data: _.extend({}, this.loadingState.get('query'), {page: nextPage, limit: BATCH_SIZE}),
 			reset: nextPage === 0,
 			remove: false,
 			success: function(collections, response) {
--- a/js/models/Bookmarks.js
+++ b/js/models/Bookmarks.js
@@ -33,12 +33,12 @@ export default Backbone.Collection.extend({
 		}
 		const nextPage = this.loadingState.get('page');
 		this.loadingState.set({page: nextPage+1, fetching: true});
-
+		var sortby = this.sortby;
 		// Show spinner after 1.5s if we're fetching a new query
 		const spinnerTimeout = setTimeout(() => nextPage === 0 && this.reset(), 1500);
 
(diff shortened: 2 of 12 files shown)


controller/lib/bookmarks.php
```diff
@@ -208,9 +208,8 @@ public function findBookmarks(
         }else{
 			$qb->selectAlias($qb->createFunction('GROUP_CONCAT(' . $qb->getColumnName('t.tag') . ')'), 'tags');
 		}
-		if (!in_array($sqlSortColumn, $tableAttributes)) {
-			$sqlSortColumn = 'lastmodified';
-		}
+		
+		$sqlSortColumn = $this->config->getUserValue($userid, 'bookmarks', 'sorting', 'lastmodified');
```
marcelklehr: It might be more flexible to call the [findBookmarks endpoint](https://github.com/nextcloud/bookmarks/blob/master/API.md#query-bookmarks) with the sort setting [from the client](https://github.com/nextcloud/bookmarks/blob/master/js/models/Bookmarks.js#L41), so that third-party clients can override the setting if they want to.

js/models/Settings.js
```diff
@@ -0,0 +1,36 @@
...
+				'requesttoken': oc_requesttoken
+			},
+			data: {
+				sorting: sorting
+			},
+			success: function() {
+				location.reload();
+			},
+		})
+	},
```
marcelklehr: You can override [Model#sync](http://backbonejs.org/#Model-sync) to streamline persisting of data. The advantage of that is that you can leverage the built in events, i.e. you could listen to settings.on('change:sorting') somewhere else and trigger a new fetch of the bookmarks, so you don't have to reload ;)

tests/settingscontroller_test.php
```diff
@@ -0,0 +1,69 @@
...
+	/**
+	 * @covers SettingsController::setSorting
+	 */
+	function testSetSorting() {
+		$output = $this->controller->setSorting('added'); //case: set a normal sorting option
+		$this->assertEquals('added', $this->config->getUserValue($this->userId,$this->appName,'sorting','')); 
+		$output = $this->controller->setSorting('foo'); //case: set an invalid sorting option
+		$data = $output->getData();
+		$this->assertEquals('error', $data['status']);
+		$this->config->deleteUserValue($this->userId, $this->appName, 'sorting'); //clean test data
```
marcelklehr: You can use a hook to clean up after the tests, like this: https://github.com/nextcloud/bookmarks/blob/master/tests/lib_bookmark_test.php#L232 ;)

tests/settingscontroller_test.php
```diff
@@ -0,0 +1,69 @@
...
+		$userManager = \OC::$server->getUserManager();
+		if (!$userManager->userExists($this->userid)) {
+			$userManager->createUser($this->userid, 'password');
+		}
+		$this->config = \OC::$server->getConfig();
+		$this->controller = new SettingsController("bookmarks", $this->request, $this->userid, $this->config);
+	}
+	
+	/**
+	 * @covers SettingsController::getSorting
```
marcelklehr: https://stackoverflow.com/questions/27605392/phpunit-trying-to-cover-or-use-not-existing-method ;)

chiotisn:
@marcelklehr I use findBookmarks from client with sortby attribute and stopped reloading page when sorting option changing. 
But there is a problem, when page loads js/apps/MainRouter fetches bookmarks before model.Settings get user's sorting option from server. 
This causes the bookmarks fetching without the sorting option at the first time and then fetching again with the sorting options (due to "listen to settings.on('change:sorting')").
I will try to fix this.

js/views/Settings.js
```diff
@@ -15,19 +16,30 @@ export default Marionette.View.extend({
...
 		'change @ui.import': 'importSubmit',
 		'load @ui.iframe': 'importResult',
-		'click .export': 'exportTrigger'
+		'click .export': 'exportTrigger',
+		'change @ui.sort': 'setSorting' 
+	},
+	initialize: function(options) {
+		this.app = options.app;
+		this.bookmarks = this.app.bookmarks;
+		this.listenTo(this.model, 'change:sorting', this.getSorting);
```
marcelklehr: This and getSorting() might be better placed in the Main App class....

js/views/Settings.js
```diff
@@ -102,5 +114,17 @@ export default Marionette.View.extend({
 	},
 	exportTrigger: function() {
 		window.location = 'bookmark/export?requesttoken='+encodeURIComponent(oc_requesttoken);
+	},
+	getSorting: function() {
+		this.getUI(this.model.get('sorting')).prop('selected',true);
+		this.bookmarks.sortby = this.model.get('sorting');
```
marcelklehr: If you use [`Collection#comparator`](http://backbonejs.org/#Collection-comparator) backbone can access the information as well ;)

marcelklehr:
@chiotisn You may trigger the loading of the settings in the Main App class and only start/create the Router after the settings have loaded.

chiotisn:
Thank you for your help! I will try to fix it! :)

js/views/Settings.js
chiotisn: Interesting!

chiotisn:
@marcelklehr I think I fixed it! :D

js/models/Bookmarks.js
```diff
@@ -33,7 +45,6 @@ export default Backbone.Collection.extend({
 		}
 		const nextPage = this.loadingState.get('page');
 		this.loadingState.set({page: nextPage+1, fetching: true});
-
 		// Show spinner after 1.5s if we're fetching a new query
 		const spinnerTimeout = setTimeout(() => nextPage === 0 && this.reset(), 1500);
```
marcelklehr: The sortby in the query would still be necessary to get word to the endpoint about the sort order, I think. :)

chiotisn: ok, I will add it again.

chiotisn: But I think that at the first query (on load) "sortby" will have no value. This happens because bookmarks are fetched before server respond to Settings fetch. I tried to fix it by loading settings at the Main and then create the Router, but, this time an EmptyBookmarks page appeared for a sec and then the bookmarks appeared sorted. Then I tried the comparator and the problem of double-fetching solved. and sorting works normally.

chiotisn:
@marcelklehr I added "sortby" attribute at the endpoint call again. I could fix the problem I have mentioned above (sortby has no value at the first request) by adding a "sorting" attribute/flag at the [loadingState](https://github.com/nextcloud/bookmarks/blob/master/js/models/Bookmarks.js#L14) set to false until Settings get sorting value from server. 
This way, fetching will happen only once and there will be a spinner on screen. However, page loading will take one more second to load. I don't think that this solution is necessary, because sorting is implemented completely by the comparator. I just mention it in case the problem with the sortby option and the endpoint is important.

marcelklehr:
@chiotisn Ah, I should have explained why sorting by comparator only will not work. As the results are broken up into pages on the server, the server will sort by date, then page, then the client will load the bookmarks and apply the comparator to sort e.g. by name. The sorting will be correct for the bookmarks displayed, but the wrong bookmarks are displayed because they're sorted after breaking after pagination.

jancborchardt:
@chiotisn @marcelklehr since it’s design-related, could you post a screenshot for easier review? :)

marcelklehr:
screenshot from 2018-06-24 18-58-06

marcelklehr (approved):
Pending @jancborchardt's approval, I say this is good to go! :)

chiotisn:
@marcelklehr thank you so much for your help!! I feel so sorry that you had to do this yourself since it was something I should have done. I had my final exams these to weeks that’s why I didn’t managed to finish it. I’m really sorry!

marcelklehr:
@chiotisn Don't fret, It wasn't too hard :) I understand. You did a great job, though!

jancborchardt:
Looks great! :) Just minor wording nitpicks:

- Call it simply "Sorting" instead of "Sorting options"
- The options themselves should only have the first word start with a capital letter:
  - Recently added (should be first cause it should be default)
  - Alphabetically
  - Most visited
  - Last modified

jancborchardt:
@chiotisn yeah, that’s fine for now. :) Feel free to do a follow-up pull request to change to Select2, but it’s fine if we first merge this. :) Small steps

jancborchardt (approved):
Fine design-wise :)

jancborchardt:
@chiotisn I also added you to the Nextcloud organization here on Github, and the Bookmarks app team. :tada: welcome!

chiotisn:
@jancborchardt Thank you very much! :)
@marcelklehr I really appreciate all your help and guidance during this period! Thanks a lot! :)

PR #504, opened 2018-06-04
Fix LinkExplorer: Always return absolute URLs

marcelklehr:


blizzz (approved):
in general i'd prefer unit tests, but this won't go without some changes: LinkPreview would need to be instantiated from a Factory that can be passed to the class constructor, so it'll be mockable. And provided phpUri does not need web access, i suppose it doesn't.

marcelklehr:
Considering that its very small, we could mock LinkExplorer and test beyond its interface?

marcelklehr:
I'll take your approval as a 'good to merge' :blush:

Issue #495, opened 2018-05-25
Improve loading of bookmarks on scroll


exekutive:
I can't stand the endless scrolling concept (where it only shows you a page of bookmarks, and doesn't load the rest until you scroll to the bottom of the window). I want to see _all_ of my bookmarks. Please, include an option to disable this. I can see how this would make sense on mobile, but there's no reason why they can't all be loaded on desktop.

marcelklehr:
I agree that it can be frustrating to have to wait for the next batch of bookmarks to load, especially when the server isn't the fastest like in my case. I do think we need pagination, though, as there are people who have more than 10,000 bookmarks. We could make the batch size configurable, though. In the latest master version, I've already [increased the batch size to 30 instead of 10 (iirc)](https://github.com/nextcloud/bookmarks/blob/master/js/models/Bookmarks.js#L5).

blizzz:
Totally agree with @marcelklehr. Perhaps we can improve the paging behaviour to trigger loading earlier, e.g. when you have scrolled just half of the height, or have a the next page already loaded, simply show it when the bottom is reached for instant presentation and a smooth feeling and have the next batch being loaded in the background already. This still might not work smooth enough when you scroll down fast. Perhaps it is just good enough.

exekutive:
No. What blizzz describes is precisely what I DON'T want. 
I would prefer to wait until the page fully loads, at least 100 bookmarks, and then see them all at once.
I don't want to have to "trigger" anything. Maybe it could be sped up with some kind of cached index.

exekutive:
Why are you asking as if my request is unusual?  It's like going to a hardware store and asking to see the hammer section and instead, the clerk insists on bringing them out to you one at a time. When I browse my files in Mac or on Windows, it doesn't wait until I scroll to show them to me. To me, THIS triggering feels weird. What if I want to jump to the end of the list? What if I feel like skimming through them?

There was nothing wrong with the way scrolling behaviour always worked. Why do you need to "improve" it? Isn't it more work to program that anyways?

 Do whatever you like, I'm just saying give me the option to use the way I'm used to.

marcelklehr:
> Why are you asking as if my request is unusual? It's like going to a hardware store and asking to see the hammer section and instead, the clerk insists on bringing them out to you one at a time.

[...]

PR #516, opened 2018-06-24
Infinite scroll: Load the first 3 pages unconditionally in sequence

marcelklehr:
as per #495

### DIFF
--- a/js/models/Bookmarks.js
+++ b/js/models/Bookmarks.js
@@ -47,6 +47,9 @@ export default Backbone.Collection.extend({
 					fetching: false,
 					reachedEnd: response.data.length < BATCH_SIZE
 				});
+				if (nextPage <= 1) {
+					that.fetchPage()
+				}
 			}
 		});
 	}


jancborchardt:
Please review @exekutive, as this is one step towards fixing what you reported at https://github.com/nextcloud/bookmarks/issues/495 :)

exekutive:
I inserted the change. It loads 30 bookmarks and stops, waits for me to scroll to the bottom and then loads 30 more.

marcelklehr:
You will need to [rebuild the app](https://github.com/nextcloud/bookmarks#manual-install), unless you insert the change in the bundled js file in `js/dist/`.

exekutive:
I changed /js/dist/main.bundle.js

marcelklehr:
Ah, my bad. Pushed a new version.

exekutive:
so I should wait for a release?

jancborchardt:
@exekutive you could manually apply the change and test it. Or are you running a local instance? (Then you could install it via git also)

marcelklehr:
The changeset in this pull request has been updated, so you can try it again (it changed from `this` to `that` ;) )

exekutive:
> you could manually apply the change and test it. Or are you running a local instance? 

> you can try it again (it changed this to that ;) )

I thought that is what I did when I changed /js/dist/main.bundle.js. It's running on a live remote server. Sorry I must not be understanding something.

The latest version I see is the one I installed 3 weeks ago (v0.12.0-rc1). Unless the only way to test this is building the app from source, in which case I'd rather just wait.

marcelklehr:
Ah, sorry foor the confusion: You correctly applied the patch, but there's a mistake in there. Instead of `this`, the new code should read `that`. I've since updated the patch here on github, but it's of course very similar to the one you tried out. So, if you simply overtake the changes as they are now, into the main.bundle.js again, it should work.

exekutive:
Got it. Thank you. Unfortunately, it still behaves the same.

marcelklehr:
Mmh. :/ I assume something went wrong applying that patch, since it works for me.
I'll merge this for now, and you can observe it in action in the next (pre-)release.

marcelklehr:
This has just hit the latest release candidate. Feel free to test and give feedback over here: https://github.com/nextcloud/bookmarks/issues/495

jancborchardt:
Good stuff, thanks a lot for your work @marcelklehr! :)

jancborchardt:
Btw @marcelklehr you are coming to the Nextcloud conference in Berlin, right? https://nextcloud.com/conf :) There’s travel/accomodation support for contributors as usual, and it would be great if you give a lightning talk (just max 5 minutes) about all the cool Bookmark app improvements! :tada:

marcelklehr:
@jancborchardt Thanks :) Regarding the conference: I won't be there, I'm afraid.

jancborchardt:
😭

PR #523, opened 2018-07-16
Feature: RSS feeds

### LINKED ISSUE #210: RSS Feed / Live bookmark
Hi there,

Like others before me, I'd really like to be able to use ownCloud bookmarks more easily, natively in my browser. This seems to be the point where usually there is a big discussion that rapidly gets technical involving Mozilla Weave and Chrome sync protocol. Here's the thing though. Really I just want to be able to do two things:
1. Add a bookmark - which is easily done throught the bookmarklette already, and...
2. Access my bookmarks - which seems to be the only part really missing.

While it would be fantastic to have a native browser plugin / addon / extension, we're never going to cover all browsers, and it seems it will take a long time to even cover the major ones. If an RSS feed could be provided though, it seems that this would be pretty simple, and in Firefox at least (and likely other browsers, I suspect) it could just be dragged to bookmarks - like the bookmarklette, to create a live bookmark, thus making all my bookmarks easily accessible.

Yes, this isn't going to provide filtering, tags etc, but it seems like a 90% solution that would require only 20% of the work of implementing full sync protocols and multibrowser addons etc.

marcelklehr:
fixes #210

- [x] Generate RSS
- [x] A link / button in the UI to easily generate the RSS URL from a bookmarks result

### DIFF
--- a/controller/rest/bookmarkcontroller.php
+++ b/controller/rest/bookmarkcontroller.php
@@ -15,6 +15,8 @@ use OCP\IL10N;
 use \OCP\IRequest;
 use \OCP\AppFramework\ApiController;
 use \OCP\AppFramework\Http\JSONResponse;
+use \OCP\AppFramework\Http\DataResponse;
+use \OCP\AppFramework\Http\TemplateResponse;
 use \OCP\AppFramework\Http;
 use \OC\User\Manager;
 use \OCA\Bookmarks\Controller\Lib\Bookmarks;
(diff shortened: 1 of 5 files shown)


Hanover96:
All bookmarks / RSS feeds are public current. Please make a function that the feeds / bookmarks can also be used private / available with login.

marcelklehr:
@Hanover96 I cannot confirm this. Could you provide a step-by-step explanation how to obtain someone's bookmarks without knowing their username and password (preferably [via email](https://github.com/marcelklehr) as this would be a serious security issue?

PR #541, opened 2018-08-03
14 compatibility

skjnldsv:
Signed-off-by: Christoph Wurst <christoph@winzerhof-wurst.at>

### DIFF
--- a/css/bookmarks.css
+++ b/css/bookmarks.css
@@ -1,13 +1,3 @@
-#content > div {
-	height: 100%;
-	overflow: auto;
-}
-
-#app-navigation {
-	height: 100%;
-	position: relative;
-}
-
 #app-navigation > ul {
 	height: auto;
 	overflow-y: hidden;
--- a/js/apps/Main.js
+++ b/js/apps/Main.js
@@ -9,7 +9,6 @@ import AppView from '../views/App';
 const Marionette = Backbone.Marionette;
 
 export default Marionette.Application.extend({
-	region: '#app-content',
 	onBeforeStart: function() {
 		var that = this;
 		this.bookmarks = new Bookmarks();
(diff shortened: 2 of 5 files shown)


js/apps/Main.js
```diff
@@ -9,7 +9,7 @@ import AppView from '../views/App';
 const Marionette = Backbone.Marionette;
 
 export default Marionette.Application.extend({
-	region: '#content',
+	region: '#app-content',
```
marcelklehr: this should probably be removed, as it's used by the AppView as a region.

js/apps/Main.js
```diff
@@ -28,7 +28,8 @@ export default Marionette.Application.extend({
 		this.router = new Router({app: this});
 	},
 	onStart: function() {
-		this.showView(new AppView({app: this}));
+		var view = new AppView({app: this});
+		view.magic();
```
marcelklehr: Might be best to make the view accessible as a property.

js/views/App.js
```diff
@@ -37,18 +36,23 @@ export default Marionette.View.extend({
 		}
 	},
 	initialize: function(options) {
+		this.bindUIElements();
```
marcelklehr: Where is this defined?

marcelklehr: Ah, that's from Marionette v2. We're using v3 ;)

css/bookmarks.css
```diff
@@ -1,13 +1,3 @@
-#content > div {
-	height: 100%;
-	overflow: auto;
-}
-
-#app-navigation {
-	height: 100%;
-	position: relative;
```
marcelklehr: mmh. I'm pretty sure that was necessary for backwards compat to nextcloud v12 ^^
However, nextcloud GmbH drops support for that when the new release is out, right? So, we probably might as well stop supporting it.

skjnldsv: Changes for 14 are pretty big. Theoritecally, we need to check, this layout should be compatible with 13 too.
So yes, we need to drop 12. :)

skjnldsv:
@marcelklehr thanks a lot for your help!! :)
I'm completely out of the loop with marionette. @ChristophWurst helped with his magic! ;)

marcelklehr:
Do my changes work on v14? I don't have an instance at hand.

skjnldsv:
@marcelklehr Yes, it looks fine here :)

PR #629, opened 2018-12-25
Add a cron job for fetching previews once a day

marcelklehr:


lib/backgroundjobs/previewsjob.php
```diff
@@ -0,0 +1,51 @@
...
+		if ($this->settings->getAppValue(
+			'core',
+			'backgroundjobs_mode'
+		) !== 'cron') {
+			return;
+		}
+		\OCP\Util::writeLog('bookmarks', 'starting PreviewsJob', \OCP\Util::WARN);
+		$allBookmarks = $this->libBookmarks->findBookmarks(-1, 0, 'lastmodified', [], true, -1);
+		foreach ($allBookmarks as $bookmark) {
+			$this->userSession->setUser($this->userManager->get($bookmark['user_id']));
```
marcelklehr: @blizzz I'm not sure about this. Are there any security implications from this? I needed it, so that the preview services can use the user cache... Or should I rather implement a dedicated app cache?

PR #718, opened 2019-05-25
Added another 3d-party client (Owncloud bookmarks)

e-alfred:
There is another 3rd-party client by Martin Jakl called "Owncloud Bookmarks". It works on all Chromium-based browsers and is very useful because it doesn't directly sync a complete folder like Floccus but allows access through a drop down list within the extension.


marcelklehr:
Thanks! :)

PR #815, opened 2019-09-24
Use routes history mode and global eslint config

skjnldsv:
- moved to regular url for better viduals
- use global eslint config
- cleaned up code
- updated makefile to fit what we use in our other apps


src/App.vue
```diff
@@ -1,22 +1,22 @@
 <template>
-  <router-view />
+	<router-view />
 </template>
 
 <script>
-import { mutations } from './store/';
+import { mutations } from './store/'
```
marcelklehr: Wut? You're doing non-semicolon style now? I like it :)

marcelklehr (approved):
Sweet, thanks alot! :)

src/App.vue
skjnldsv: Yes, we removed it a few time ago in the default eslint config :)
When I started migrating things to vue, then everyone just used my config, so in the end almost all newly migrated apps are using similar (very close) configs. Thus the new unified config ;)

skjnldsv:
:tada: :tada: :tada:

PR #842, opened 2019-10-26
Fix some translation problems

PFischbeck:
In this PR, I made some strings translatable.
Additionally, I changed some whitespace for several other translations, since the [translation tool](https://github.com/nextcloud/docker-ci/blob/8daabc1bdf7a37af01f7dc0e7cc2a4a68465b1f1/translations/translationtool/src/translationtool.php) doesn't recognize them if there is whitespace between the opening bracket and the first quotation mark. I'm also working on a PR to allow for whitespace there.

### DIFF
--- a/src/components/BookmarksList.vue
+++ b/src/components/BookmarksList.vue
@@ -21,8 +21,8 @@
 		<div
 			v-else-if="!loading && !folderChildren.length"
 			class="bookmarkslist__empty">
-			<h2>No bookmarks here</h2>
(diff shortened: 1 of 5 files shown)


marcelklehr:
Awesome! Thank you so much for stepping up and fixing these! :rocket: :stars: :)

Most of these are due to code auto-formatting, which will be hard to fix long-term, but as you've already opened a PR on the translation tool, this is great!

PFischbeck:
For reference, the PR for the translation tool can be found here: nextcloud/docker-ci#177

PR #854, opened 2019-11-10
Make folders clickable

manuee:
### Problem
Currently in order to navigate bookmark folders you have to click on the title of the folder, which is not ideal. Users expect to be able to click on the folder itself to open it, which is how files folders work.

### Propose solution
Have the whole icon be clickable.

Note that a similar argument could be made for the bookmarks themselves, although there perhaps one could expect to click on the URL itself, so sticking to just folders on this PR.

### DIFF
--- a/src/components/Folder.vue
+++ b/src/components/Folder.vue
@@ -1,10 +1,11 @@
 <template>
-	<div :class="{folder: true, 'folder--gridview': viewMode === 'grid'}" @click="onSelect">
-		<figure class="folder__icon icon-folder" />
+	<div :class="{folder: true, 'folder--gridview': viewMode === 'grid'}">
+		<figure class="folder__icon icon-folder" @click="onSelect" />
 		<template v-if="!renaming">
 			<h3
 				class="folder__title"
-				:title="folder.title">
+				:title="folder.title"
+				@click="onSelect">
 				{{ folder.title }}
 			</h3>
 			<Actions class="folder__actions">
@@ -93,7 +94,6 @@ export default {
 	display: flex;
 	align-items: center;
 	position: relative;
-	cursor: pointer;
 }
 
 .folder:hover {
@@ -106,6 +106,7 @@ export default {
 	width: 20px;
 	background-size: cover;
 	margin: 15px;
+	cursor: pointer;
 }
 
 .folder--gridview .folder__icon {
@@ -123,6 +124,7 @@ export default {
 	text-overflow: ellipsis;
 	overflow: hidden;
 	white-space: nowrap;
+	cursor: pointer;
 	margin: 0;
 	padding: 15px 0;
 }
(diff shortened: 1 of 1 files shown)


marcelklehr:
Awesome! Thanks for stepping up to do this :)

Is the three-dot-menu still functioning with this change?

manuee:
Thanks for the quick reply!

> Awesome! Thanks for stepping up to do this :)
> 
> Is the three-dot-menu still functioning with this change?

I've not actually tested this, (I'm very new to Nextcloud development this is my very first try!)
All I have is a local Nextcloud instance running via the snap (on Ubuntu). I tried updating the code within the snap, but apparently it didn’t have any effect. In other words, I don’t know how to test this :smile: 

I noticed that the CSS would need updating as well to make the cursor a pointer on the whole icon so did that in the last commit.

marcelklehr:
You will need [to run the build script](https://github.com/nextcloud/bookmarks#manual-install) after changing the client-side code.

manuee:
OK after some blood and tears I managed to set this up locally to test it, outside of a snap context.

1. The  three-dot-menu is clickable, but wehn you click it you also click the div so you open the folder instead of the menu.
2. The css isnt working as I would expect, the pointer only shows up if your not on the folder title or the folder icon.

So this is not as simple as I thought it would be, seems like this would require changing the markup / CSS etc of the folder component, and I'm not familiar enough with Vue to tackle this...

manuee:
Seems to me thought like we should replicate what files does perhaps? Anyway, thanks for having a look!

src/components/Folder.vue
```diff
@@ -1,11 +1,10 @@
 <template>
-	<div :class="{folder: true, 'folder--gridview': viewMode === 'grid'}">
+	<div :class="{folder: true, 'folder--gridview': viewMode === 'grid'}" @click="onSelect">
```
marcelklehr: try `@click.stop`

marcelklehr: ah, wrong spot.

marcelklehr: You can check the event.currentTarget in `onSelect() {`

manuee: `event.currentTarget` is the div itself, even when clicking on the dotted menu unfortunately

marcelklehr: Ah, it's probably only `target`

marcelklehr: I'm also in IRC, btw ;)

manuee:
OK I gave this a bit of thought, and I think I figured out a less disruptive change, making the folder icon clickable instead of the whole div (and keeping the title clickable as well).

Peek 2019-11-11 15-46

Should be an improvement to the users as they probably will aim for the icon anyway, and we don’t have to mess about with event targets, or re-factoring a big chunk of the component.

marcelklehr (approved):
Sweet. Thanks a lot :)

manuee:
Brilliant, thanks!

PR #906, opened 2020-01-29
Fix tag and count display in navigation sidebar.

iamleeg:
The problem we saw was that the tag names and counts weren't showing up. By inspecting the state object, we saw that the tag name and count properties returned from the backend were called `tag` and `nbr`.

Signed-off-by: Graham Lee <leeg@labrary.online>

### DIFF
--- a/lib/Db/TagMapper.php
+++ b/lib/Db/TagMapper.php
@@ -33,14 +33,14 @@ class TagMapper {
 	public function findAllWithCount($userId) {
 		$qb = $this->db->getQueryBuilder();
 		$qb
-			->select('t.tag')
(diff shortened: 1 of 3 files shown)


marcelklehr:
Thank you for investigating this. I'm in the process of refactoring the backend in the master branch, which is probably why the labels are off. However, as the refactoring shouldn't change the output of the API, the place to fix this would be the backend ;)

iamleeg:
Great, thanks! Are you doing that or do you want us to look into it?

marcelklehr:
I won't be touching the TagMapper in the next steps, so I'd be happy if you would look into this. Help is always appreciated :)

marcelklehr:
:tada: Thanks folks! :)

PR #914, opened 2020-02-07
Update error handling on "Import"

funwhilelost:
The test for `response.ok` doesn't appear to work in latest Brave or Chrome.  The result is that, although the POST to `/import` succeeded, it throws an error with `[Object object] [Object object] ...`.  This test ensures that errors are thrown only when `response.data.success` is not set correctly.  The output of `response.data.data.join('\n')` was worthless in the case of my bug so I wasn't sure which style should be used on actual error scenarios.  Please advise.

image

### DIFF
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -207,16 +207,12 @@ export default {
 		return axios
 			.post(url(`/bookmark/import`), data)
 			.then(response => {
-				if (!response.ok) {
+				if (!response.data || !response.data.status === 'success') {
 					if (response.status === 413) {
 						throw new Error('Selected file is too large')
 					}
-					throw new Error(response.data.data.join('\n'))
-				} else {
-					const status = response.data.status
-					if (status !== 'success') {
-						throw new Error(response.data)
-					}
+					throw new Error(response.data)
+					console.error('Failed to import bookmarks', response)
 				}
 			})
 			.catch(err => {


src/store/actions.js
```diff
@@ -207,16 +207,12 @@ export default {
...
 					if (response.status === 413) {
 						throw new Error('Selected file is too large')
 					}
 					throw new Error(response.data.data.join('\n'))
-				} else {
-					const status = response.data.status
-					if (status !== 'success') {
-						throw new Error(response.data)
-					}
+					// or throw new Error(response.data)??
```
marcelklehr: We could log the data  to the console and display a generic error message, pointing the user to the console for more information.

funwhilelost: Sounds good.  I found a number of other examples of `throw new Error(response.data)` so I went with that convention and added a `console.error` report with more details.

marcelklehr: Sweet. Does this improve the error experience in your opinion? (I'm not sure what this would output now)

funwhilelost: @marcelklehr I don't have a good repro for an actual failure from the back-end API so I'm not certain what the error experience would be but it does fix the false-positive on the failure condition.

marcelklehr: Ohh, I think I might have missed the point of this PR :sweat_smile: 

So, in general the `status=success/error` property is not directly tied to the response code (even though this would be desirable). Still, this is a good improvement of the code.

PR #952, opened 2020-03-28
Sharing UI

marcelklehr:
Sharing user:
Screenshot from 2020-03-31 14-39-24
Screenshot from 2020-03-31 14-40-39

Sharee:

Screenshot from 2020-03-31 14-40-00

Public link
Screenshot from 2020-03-31 14-43-03


jancborchardt (approved):
Wow, this is very very cool! My only feedback is kinda unrelated. :D
- On the sharing page, the title subline is black but should be white
- Instead of breadbrumbs, we should use the system that Photos uses with the nice large headings

So :+1: from me

skjnldsv:
> * Instead of breadbrumbs, we should use the system that Photos uses with the nice large headings

I think there is an issue already.
Let's stay relevant to this issue ;)

jancborchardt:
> > * Instead of breadbrumbs, we should use the system that Photos uses with the nice large headings
> 
> I think there is an issue already.
> Let's stay relevant to this issue ;)

Exactly why I simply gave my :+1: and said it’s unrelated. ;)

marcelklehr:
Ah, nice I hadn't seen the photos app approach, yet :)

> On the sharing page, the title subline is black but should be white

I was baffled by that as well. I'm using the PublicPageTemplate to set those, so I wasn't sure whether I should style that myself, or if this is an upstream bug.

skjnldsv:
Please rebase ;)
Tests should be running now, I feel more confident reviewing this pr now :)

marcelklehr:
The server-side tests are broken, atm, I will work on this laters

marcelklehr:
I should get those complexity numbers down... :sweat_smile:

jancborchardt:
So @skjnldsv will be able to talk about this more, but the Sharing UI would be a good candidate for a component in the Vue components library :) https://github.com/nextcloud/nextcloud-vue

marcelklehr:
@skjnldsv This is good to review now :)

ccoenen:
`omg-its-happening.gif` here. Thanks everyone! Looking forward to the release!

jancborchardt:
Props @marcelklehr, really good work! :)

PR #963, opened 2020-04-07
actions.js file fix

Valdnet:
Adding a space character.


marcelklehr:
Ah, thanks for the heads-up! I realized that I need to properly parameterize the messages...

Issue #778, opened 2019-08-25
Feature request: retain bookmark order


InfamousUser:
Would be nice to have the bookmarks retain the order when syncing with Floccus.
I have my bookmarks somewhat sorted by relative order of the bookmarks. When reviewing the bookmarks on the server I do not see a way to view them without being resorted (desorted).

weeman1337:
Does anybody know whether Floccus sends the order to Nextcloud?

marcelklehr:
Yes, the order is stored in nextcloud bookmarks, it's just not displayed or possible to change in the web UI.

weeman1337:
Thanks @marcelklehr . I assume it is the `index` column of the `bookmarks_tree` table, right?

If I click on a folder in the front-end I am not able to find the request and response of the item list. Could you give me a hint?

![image](https://user-images.githubusercontent.com/6216686/85121651-708c5300-b225-11ea-829b-0459c97ed38b.png)

marcelklehr:
Currently the UI uses two older endpoints, namely `bookmark` and `folder`. There is also a childorder endpoint that gives you only the order and the ids without any content fields, and finally, there's the children endpoint that gives you the contents in order. All documented over here: https://nextcloud-bookmarks.readthedocs.io/en/latest/

I've neglected implementing this so far because I felt, if the order was displayed in the UI, it should also be possible to modify it, but implementing *that* is probably quite a hassle. I'm also happy with just a PR for only displaying in order, should you want to work on this :)

weeman1337:
> I'm also happy with just a PR for only displaying in order, should you want to work on this :)

Agree. Lets start with this. The current use-case on my side is sync bookmarks between browsers. Bonus: Online access via Nextcloud. But at the moment it is annoying using the web app because I do not find the bookmarks in my bookmark jungle :D

Any opinions about how to name the order option?
* _Original order_
* _Source order_

marcelklehr:
Maybe *custom order*? :man_shrugging:

InfamousUser:
I'm also trying to have bookmark sync between browsers, but if they don't sort the same way and it's not possible to change order it's not of use. We need Chrome-like bookmark sync! But I don't want to send my bookmarks to Google.

marcelklehr:
To be clear: The bookmarks app does store the bookmarks order for syncing. It's just not displayed in the UI, yet.

InfamousUser:
[...]

PR #1092, opened 2020-06-22
Add custom order display option

weeman1337:
* Fix contribute link
* Add the new *custom order* option
* Add support for the new sort setting
* Tweak the bookmark controller to support sort by `index`

I really do not see any chance me rewriting the app to the bookmarks API v2. So I extended the bookmark controller and added support for the new sort option.

Close issue #778


weeman1337:
@marcelklehr could you send me info on about how to set up the translation for the new string? I could also provide the first translation.

marcelklehr:
New strings are automatically picked up by the transifex bot and translatable via transifex.

marcelklehr:
Sorry for the late reply. I'm not sure if I like this approach, as the server actually also stores folder's positions and the bookmarks endpoint is also used for other queries where sorting by index doesn't make sense.

> I really do not see any chance me rewriting the app to the bookmarks API v2.

The bookmarks UI already uses the v2 API, but it's using the private routes not the public routes. I probably should have explained that.

weeman1337:
> I'm not sure if I like this approach

@marcelklehr how would you like it to be?

marcelklehr:
I was thinking of using this endpoint: https://nextcloud-bookmarks.readthedocs.io/en/latest/folder.html#get-folder-s-contents

But that likely requires some more refactoring.

weeman1337:
> I was thinking of using this endpoint: https://nextcloud-bookmarks.readthedocs.io/en/latest/folder.html#get-folder-s-contents

This endpoint doesn't support sorting and the returned object do not contain all the fields for doing it in the front end (index, number of clicks, creation time).

weeman1337:
ping @marcelklehr

marcelklehr:
> ping

Sorry :see_no_evil: 

> This endpoint doesn't support sorting and the returned objects do not contain all the fields for sorting the front end (index, number of clicks, creation time).

This endpoint already returns the ordered data, but is indeed missing some items.

> The order could be retrieved by GET /public/rest/v2/folder/(int: folder_id)/childorder but then still number of clicks and creation time is missing.

This endpoint is supposed to only provide the order while the actual data comes from the /folder and /bookmark endpoints.

I think a good way to implement "custom ordering" would be something like this

1) [if ordering != 'custom'] display bookmarks and folders as returned from the API with `sortby` param
2) [if ordering == 'custom'] fetch childorder endpoint for currently active folder and arrange already loaded bookmarks and folders according to that order.

What do you think?

weeman1337:
Thanks for the answer.

> What do you think?

It is kind of inconsistent, but works. For an API client it would be much easier to pass the desired order as a query param to a single API endpoint.

> This endpoint already returns the ordered data, but is indeed missing some items.

It returns the "custom order", yes. But it is not possible to request a different one.

marcelklehr:
Yeah the reason for this inconsistent design, is that the "custom order" is kind of part of the data model to be consistent with browser implementations, while all other possible orderings are just "ORDER BY" appendices on the bookmarks table. This is because the custom order explicitly "includes" folders, so you can e.g. sort 3 bookmarks to the top, then have two folders and then the rest of the bookmarks, which is not possible with the normal sort options (those are bookmarks-only).

marcelklehr:
I've finally had some time to mull this over in my head and I think your approach is viable together with a small change in the way the results are displayed that I've committed to your branch just now. Let me know what you think :)

marcelklehr (approved):

weeman1337:
> I've finally had some time to mull this over in my head and I think your approach is viable together with a small change in the way the results are displayed that I've committed to your branch just now. Let me know what you think :)

Thanks. Go from my side ✓

src/components/BookmarksList.vue
```diff
@@ -7,19 +7,42 @@
 		<CreateBookmark v-if="newBookmark" />
 		<CreateFolder v-if="newFolder" />
 		<template v-if="$route.name === routes.FOLDER || $route.name === routes.HOME">
-			<Folder
-				v-for="folder in folderChildren"
-				:key="'f' + folder.id"
-				:folder="folder" />
+			<!-- FOLDER VIEW WITH CUSTOM SORTING -->
+			<template v-if="$store.state.settings.sorting === 'index'">
```
weeman1337, line 12: Wouldn't it be better to sort somewhere in the "frontend backend" ;)  
Like by using a getter of the store.

src/components/BookmarksList.vue
```diff
@@ -7,19 +7,42 @@
...
+				<template v-for="item in children">
+					<Folder
+						v-if="item.type === 'folder'"
+						:key="item.type + item.id"
+						:folder="$store.getters.getFolder(item.id)[0]" />
```
weeman1337, line 17: In my opinion the store shouldn't be accessed from the template.

PR #1137, opened 2020-08-01
Implement unified search provider

marcelklehr:
see https://github.com/nextcloud/server/pull/21971

- [x] deal with nc<20 compatibility
- [x] deal with in-app search
- [x] implement a proper route for individual bookmarks


skjnldsv:
> Bookmarks has its own search currently. Any recommendations on how this should be handled?

Inner filtering can be done on the top right search design, like Talk.
For the global search you can rely on the Unified search.

But regarding bookmarks, I'm wondering if this shouldn't be done with Unified search only, no?

src/components/Breadcrumbs.vue
```diff
@@ -224,67 +222,67 @@ export default {
...
-	padding: 0;
+  flex: 2;
+  display: flex;
+  flex-direction: row-reverse;
+  padding: 0;
 }
 
 .breadcrumbs__controls > * {
-	min-width: 30px;
+  min-width: 30px;
```
skjnldsv, line 286: Tabs? :p

marcelklehr: Ooops, github only displayed your comments only after I merged. Ah will fix those tabs.

marcelklehr:
> But regarding bookmarks, I'm wondering if this shouldn't be done with Unified search only, no?
You only have one type of items to search in the end :)

True, but I want to support older nextcloud versions as well, so moving the search into the control bar is a good solution, yeah.

skjnldsv:
> True, but I want to support older nextcloud versions as well, so moving the search into the control bar is a good solution, yeah.

You can still use the old search with a simple class check to see if the UnifiedSearch is enabled and active :shrug: 
The old standard was `OCA.Search`, you were already using it?
If so, just keep both and only register your `OCA.Search` provider if the class exists after page is done loading?

That would keep it simple and avoid you using your own input?

marcelklehr:
Ah, nice idea.Thanks :)

skjnldsv:
https://github.com/nextcloud/server/pull/9912

PR #1138, opened 2020-08-02
Implement Flow integration

marcelklehr:
@blizzz I'd like to add some events here, too, but I'm kinda at a loss on how I would go about that. I already expose events on the global dispatcher. Do I merge those with the flow events? Or are those different events? What would be the event subject? Can you guide me through this a little?

### DIFF
--- a/lib/Flow/CreateBookmark.php
+++ b/lib/Flow/CreateBookmark.php
@@ -5,19 +5,23 @@ namespace OCA\Bookmarks\Flow;
 
 
 use OCA\Bookmarks\Service\BookmarkService;
+use OCA\WorkflowEngine\Entity\File;
 use OCP\EventDispatcher\Event;
-use OCP\EventDispatcher\GenericEvent;
 use OCP\EventDispatcher\IEventDispatcher;
+use OCP\Files\Folder;
+use OCP\Files\Node;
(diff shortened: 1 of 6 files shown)


lib/Flow/CreateBookmark.php
```diff
@@ -0,0 +1,167 @@
...
+	 */
+	public function getIcon(): string {
+		return $this->urlGenerator->imagePath('bookmarks', 'bookmarks.svg');
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function isAvailableForScope(int $scope): bool {
+		return true;
```
blizzz, line 96: this also allows administrator to configure global rules for creating bookmarks. Is this intended?

marcelklehr: Yeah, I wasn't sure what the use case for this would be, but I thought it couldn't hurt.

blizzz: the parameter is one of the scope constants from IManager, `SCOPE_ADMIN` or `SCOPE_USER`. Some operations might be available for both scopes, others like running arbitrary scripts only for admins, because they could do anything, and others like posting to Talk is personal only as it is bound to the user who configured the rule.

PR #1146, opened 2020-08-07
Implement dead link detection

marcelklehr:
fixes #1132

PR #1147, opened 2020-08-07
Design upgrades

marcelklehr:
- [ ] use md icons where appropriate
- [ ] use nextcloud breadcrumbs
- [x] fix folder icons by using mdi with fillColor primary
- [x] rebase onto  #1146 and add icon for that


jancborchardt:
> use nextcloud breadcrumbs

We basically phased out breadcrumbs in any app except Files (e.g. in Deck and Forms) cause in the other apps, if we have a hierarchy, it’s usually in the AppNavigation. Any specific reason to keep it? :)

marcelklehr:
> We basically phased out breadcrumbs in any app except Files (e.g. in Deck and Forms) cause in the other apps, if we have a hierarchy, it’s usually in the AppNavigation. Any specific reason to keep it? :)

I see. We've had a similar conversation before, I think :D In bookmarks the hierarchy used to live in the app navigation for a while but it was not ideal, because users can create an arbitrarily deep hierarchy (which is necessary for browser compatibility), so it's not a two-tier hierarchy like in other apps. So, I think a UI similar to files makes sense.

PR #1359, opened 2020-12-22
add qutebrowser userscript

geeseven:
Recently, qutebrowser added a userscript for adding bookmarks for this app. Let me know if this is too niche.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ This app provides you with a web interface for collecting and organizing bookmar
 - [Owncloud Bookmarks](https://chrome.google.com/webstore/detail/owncloud-bookmarks/eomolhpeokmbnincelpkagpapjpeeckc?hl=de) - Bookmarks extension for Chromium-based browsers (Chromium/Chrome/Opera/Vivaldi)
 - [Floccus](https://github.com/marcelklehr/floccus) - Bookmark sync for Firefox/Chromium-based browsers
 - [FreedomMarks](https://github.com/damko/freedommarks-browser-webextension) - Addon for Firefox and Chrome. No sync, just a client.
+- [add-nextcloud-bookmarks](https://github.com/qutebrowser/qutebrowser/blob/master/misc/userscripts/add-nextcloud-bookmarks) - qutebrowser userscript that allows for easy bookmark creation
 
 ### Desktop
 - [Nextcloud Bookmark Manager](https://www.midwinter-dg.com/mac-apps/nextcloud-bookmark-manager.html) A MacOS client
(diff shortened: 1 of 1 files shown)


marcelklehr:
Cool! How would people use that?  (I've never used qutebrowser.) Is there a better link than the source code?

geeseven:
I would consider qutebrowser to be used mainly by 'advanced users', as it has vim like keyboard controls with a minimal GUI.  The userscripts are a way for users to add basic functionality, via self contained scripts, until a full blown plugin system is added.  Also, I would consider userscripts for 'advanced qutebrowser users'.  Unfortunately, there is not a centralized location for userscript documentation.  It is spread over [Configuring qutebrowser], [Documentation of settings], and even [How to write userscripts].

The comment block at the top of the script gives information on how to use the script. Since there has not been a release since the script was merged, a user would need to add it to `~/.local/share/qutebrowser/userscripts/` and make it executable.  Once a new release is added, this won't be needed.  Personally, I added `config.bind('X', 'spawn --userscript add-nextcloud-bookmarks')` to `~/.config/qutebrowser/config.py`.  This way hitting X will bookmark the current URL.

Since this is a self contained script, there is not a fancy website or documentation page.

If this is too niche, I would understand if you choose not to merge this pull request.


[Configuring qutebrowser]: https://www.qutebrowser.org/doc/help/configuring.html
[Documentation of settings]: https://www.qutebrowser.org/doc/help/settings.html
[How to write userscripts]: https://www.qutebrowser.org/doc/userscripts.html

marcelklehr:
> If this is too niche, I would understand if you choose not to merge this pull request.

I don't have a policy of only adding mainstream stuff. I'm not sure if linking the source code is helpful, though, if this will be shipped with the next release. If there was some kind of list of userscripts in the docs or a guide on how to add it, that would be nice.

geeseven:
>  I'm not sure if linking the source code is helpful, though, if this will be shipped with the next release.

That is a good point.  


> If there was some kind of list of userscripts in the docs or a guide on how to add it, that would be nice.

[Here] is a list of included (in source code) and 3rd party userscripts.  The 'Writing qutebrowser userscripts' section of [How to write userscripts] describes how to add userscripts.

Maybe we should wait till the next release of qutebrowser? :shrug: Then have something like:

- qutebrowser includes add-nextcloud-bookmarks userscript which allows for for easy bookmark creation.  See [comment] for configuration details.


[Here]: https://github.com/qutebrowser/qutebrowser/blob/master/misc/userscripts/README.md
[How to write userscripts]: https://www.qutebrowser.org/doc/userscripts.html
[comment]: https://github.com/qutebrowser/qutebrowser/blob/master/misc/userscripts/add-nextcloud-bookmarks

marcelklehr:
I think there's been a release now, no?

Issue #1369, opened 2021-01-07
bookmarks cache folder is filled up with 4 byte files


BigMichi1:
**Describe the bug**
the ``bookmarks/cache`` folder in the ``appdata`` directory contains hundreds of files with a 4 byte size and when inspecting the content only ```null``` is written there.

**To Reproduce**
1. configure pageres as preview engine
2. clear cache folder by manually removing the entire content
3. reload oropen a bookmarks folder containing some links
4. many 4 byte files are stored

**Expected behavior**
these ``null`` 4 byte files should not be stored as there are eating the inodes and also according to the file name they belong to the not configured preview engines

**Screenshots**
![image](https://user-images.githubusercontent.com/91786/103915219-34f69f00-510b-11eb-94a3-f1a97b2f2f7b.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Edge
 - Version: 87.0.664.66

**Server (please complete the following information):**
 - OS: Ubuntu 20.04.1
 - HTTP server: nginx
 - Database: PostgreSql 12.4
 - PHP version: 7.4.13
 - Nextcloud version: 20.0.4
 - Bookmarks app version: 4.0.5

marcelklehr:
Fixed by a8cda09

PR #1387, opened 2021-01-23
Fix nc21 issues

marcelklehr:
Signed-off-by: Marcel Klehr <mklehr@gmx.net>

fixes #1404

PR #1400, opened 2021-02-01
Run GC on cache dir and add CACHDIR.TAG

### LINKED ISSUE #1370: tagging cache folder as a cache by adding a CACHEDIR.TAG file
**Is your feature request related to a problem? Please describe.**
currently when the appdata folder from nextcloud is included in a backup also the cache folder gets included, even if it only contains data that does not need to be included.

**Describe the solution you'd like**
tagging the cache folder created by the bookmarks app can help reduce the size of backups by adding one additional file named ``CACHEDIR.TAG`` containing something similar to
```
Signature: 8a477f597d28d172789f06886806bc55
# This file is a cache directory tag created by the nextcloud bookmarks app.
# For information about cache directory tags, see:
#       http://www.brynosaurus.com/cachedir/
```

**Describe alternatives you've considered**
exclusion when backups are specified can be very complicated as there are some parts in the complete path that are random

**Additional context**
the CACHEDIR.TAG file is used in many applications like fontconfig or man to exclude these directories from backups. Also the file is recognized by many backup solutions out there and compressors like tar.
specification can be found here https://bford.info/cachedir/

marcelklehr:
fixes #1370 and #1369

### DIFF
--- a/lib/BackgroundJobs/CrawlJob.php
+++ b/lib/BackgroundJobs/CrawlJob.php
@@ -7,7 +7,8 @@
 
 namespace OCA\Bookmarks\BackgroundJobs;
 
-use OC\BackgroundJob\TimedJob;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\TimedJob;
 use OCA\Bookmarks\Db\Bookmark;
 use OCA\Bookmarks\Db\BookmarkMapper;
 use OCA\Bookmarks\Service\BookmarkPreviewer;
@@ -40,10 +41,15 @@ class CrawlJob extends TimedJob {
 	 * @var CrawlService
 	 */
 	private $crawler;
+	/**
+	 * @var ITimeFactory
+	 */
+	private $timeFactory;
 
 	public function __construct(
-		IConfig $settings, BookmarkMapper $bookmarkMapper, CrawlService $crawler
+		IConfig $settings, BookmarkMapper $bookmarkMapper, CrawlService $crawler, ITimeFactory $timeFactory
 	) {
+		parent::__construct($timeFactory);
 		$this->settings = $settings;
 		$this->bookmarkMapper = $bookmarkMapper;
 
--- a/lib/Service/FileCache.php
+++ b/lib/Service/FileCache.php
@@ -24,6 +24,18 @@ class FileCache implements ICache {
 			$appData->newFolder('cache');
 			$this->storage = $appData->getFolder('cache');
 		}
+		if (!$this->storage->fileExists('CACHEDIR.TAG')) {
+			try {
+				$this->storage->newFile('CACHEDIR.TAG',
+					'Signature: 8a477f597d28d172789f06886806bc55' . "\r\n" .
+					'# This file is a cache directory tag created by the nextcloud bookmarks app.' . "\r\n" .
+					'# For information about cache directory tags, see:' . "\r\n" .
+					'#       http://www.brynosaurus.com/cachedir/)' . "\r\n"
+				);
+			} catch (NotPermittedException $e) {
+				// No op
+			}
+		}
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


marcelklehr:
@stijink Coukd you have a look over this as well. Static analysis is failing, but I'm fixing that with #1387 Thanks :)

lib/BackgroundJobs/CrawlJob.php
```diff
@@ -40,10 +41,15 @@ class CrawlJob extends TimedJob {
 	 * @var CrawlService
 	 */
 	private $crawler;
+	/**
+	 * @var ITimeFactory
+	 */
+	private $timeFactory;
```
stijink, line 47: @marcelklehr What's the minimum php version required for nextcloud ?

lib/Service/FileCache.php
```diff
@@ -24,6 +24,18 @@ public function __construct(IAppData $appData) {
 			$appData->newFolder('cache');
 			$this->storage = $appData->getFolder('cache');
 		}
+		if (!$this->storage->fileExists('CACHEDIR.TAG')) {
+			try {
+				$this->storage->newFile('CACHEDIR.TAG',
+					'Signature: 8a477f597d28d172789f06886806bc55' . "\r\n" .
```
stijink, line 30: If this signature is used in other places than this function you should consider to use a variable or constant.

tests/BackgroundJobTest.php
```diff
@@ -54,9 +60,18 @@ protected function setUp() :void {
 	 * @doesNotPerformAssertions
 	 */
 	public function testPreviewsJob() {
+		$this->previewsJob->setId(1);
 		$this->previewsJob->execute($this->jobList);
 	}
 
+	/**
+	 * @doesNotPerformAssertions
```
stijink, line 68: Maybe you can explain the purpose of this test. You "just" use it to execute the function ? You "could" use mocking here to make the test more useful.

lib/Service/FileCache.php
```diff
@@ -24,6 +24,18 @@ public function __construct(IAppData $appData) {
 			$appData->newFolder('cache');
 			$this->storage = $appData->getFolder('cache');
 		}
+		if (!$this->storage->fileExists('CACHEDIR.TAG')) {
```
stijink, line 27: The thing that comes to my mind is why do you as a "plugin developer" need to care about such things a tagging your cache ? From my perspective this would be a nice thing to have from the nextcloud src directly. Maybe by implementing an abstract class.

stijink (approved):

lib/BackgroundJobs/CrawlJob.php
marcelklehr: I think nc21 will require a minimum of 7.4, v4.x of the bookmarks app has a min-requirement of 7.3. Nc20 has a min-requirement of 7.2 or 7.3 I think, so this way we're still compatible with v20, but I'm looking forward to using the new features, though :) Also, we're already compatible with php8 thanks to #1387

lib/Service/FileCache.php
marcelklehr: It probably won't as it's only there so that backup programs don't backup the folder. A comment would probably be good, though :D

lib/Service/FileCache.php
marcelklehr: Yeah, Nextcloud only provides a memory cache implementation. That's why I created the file cache implementation myself.

tests/BackgroundJobTest.php
marcelklehr: Yeah, it's a rather sloppy test, I agree. I'll try to look into how I can make it more meaningful.

lib/Service/FileCache.php
stijink: https://github.com/mockery/mockery is a rather nice tool for those kind of tests. It integrates seemlessly into phpunit

tests/BackgroundJobTest.php
marcelklehr: I've used phpunit's stubbing functionality to improve these tests in the latest commit :)

PR #1401, opened 2021-02-01
API: Fix renaming and deleting tags

### LINKED ISSUE #1334: 405 on PUT + DELETE for /public/rest/v2/tag/<TAG>
**Describe the bug**
Deleting or renaming tags via the public REST API will result in a 405

**To Reproduce**
DISCLAIMER: Add your credentials to these curls

***DELETE***
`curl --location --request DELETE 'https://your.cloud.example/index.php/apps/bookmarks/public/rest/v2/tag/deleteMe'`

***PUT***
`curl --location --request PUT 'https://your.cloud.example/index.php/apps/bookmarks/public/rest/v2/tag/deleteMe' \
--data-raw '{ "name": "somethingElse" }'`

**Expected behavior**
[API Doc](https://nextcloud-bookmarks.readthedocs.io/en/latest/tag.html)

**Desktop:**
 - OS: MacOS X

**Server:**
 - Database: mysql 10.3.27
 - PHP version: 7.4.11
 - Nextcloud version: 19.0.3
 - Bookmarks app version: 3.4.9
 - Nextcloud external user backend: none

marcelklehr:
fixes #1334


marcelklehr:
@stijink Could you review this? :)

appinfo/routes.php
```diff
@@ -92,9 +92,9 @@
 		['name' => 'tags#full_tags', 'url' => '/public/rest/v2/tag', 'verb' => 'GET'],
 		['name' => 'tags#rename_tag', 'url' => '/public/rest/v2/tag', 'verb' => 'POST'],
 		['name' => 'tags#delete_tag', 'url' => '/public/rest/v2/tag', 'verb' => 'DELETE'],
-		['name' => 'tags#rename_tag', 'url' => '/tag/{old_name}', 'verb' => 'POST'],
-		['name' => 'tags#rename_tag', 'url' => '/tag/{old_name}', 'verb' => 'PUT'],
-		['name' => 'tags#delete_tag', 'url' => '/tag/{old_name}', 'verb' => 'DELETE'],
+		['name' => 'tags#rename_tag', 'url' => '/public/rest/v2/tag/{old_name}', 'verb' => 'POST'],
```
stijink, line 95: @marcelklehr Seems about right for me. But just out of interest. Maybe you can explain to me the syntax behind this convention. Is this the public API specially of the bookmark app, or is this a nextcould-wide convention ?

And another question about the routing in general. Is this a routing coming from a framework or is this a nextcloud-thing ?

stijink (approved):

appinfo/routes.php
marcelklehr: The paths in `routes.php` will be appended to `/apps/bookmarks`. AFAIK the routing is currently a nextcloud thing, perhaps they use a symfony implementation behind the scenes, but it's not exposed. Generally, Nextcloud tries to only expose and make use of their own interfaces (in the OCP namespace), even if they use Symfony or other stuff behind the scenes (with the exception of PSR interfaces, which are used directly, afaict).

The `/public/rest/v2` path is bookmarks specific. The whole API is documented over here: https://nextcloud-bookmarks.readthedocs.io/en/latest/ (The internal API is largely identical with the difference of cookie auth being used and CSRF being applied.)

stijink: @marcelklehr You could use private methods to generate public / private Routes. This way it should be less error-prone.

eg.
```php
private function publicRoute(string $name, string $url, string $verb)
{
    return [
        'name' => $name,
        'url' => '/public/rest/v2/' . $url,
        'verb' => $verb,
    ]
}
```

marcelklehr: Ah, that's a nice idea, indeed!

Issue #681, opened 2019-04-11
Add recycle bin to Bookmarks UI to make restoring deleted bookmarks easier


e-alfred:
### Steps to reproduce
1. Delete bookmark in Bookmarks app

### Expected behaviour
Bookmarks should be put into a recycle bin for easy restoration or final deletion. This would prevent deleting bookmarks accidentally and users would be able to restore them easily. Also, a deletion from a client by accident would make restoration easier.

### Actual behaviour
Bookmarks are deleted without the ability to restore them.

DeathByDenim:
Hi @marcelklehr ,

I got hit by the lack of a way to undelete bookmarks after Floccus deleted half of my bookmarks due to syncing troubles.

Would you be ok with me having a stab at this? I've been tinkering with the Bookmarks app a bit already and got it to display a "Deleted bookmarks" menu entry similar to the Files app. I also managed to give bookmarks a "deleted" property.
![image](https://user-images.githubusercontent.com/5130183/103491340-19308780-4df1-11eb-9fa9-6e4948d86eeb.png)

Let me know!

marcelklehr:
@DeathByDenim That would be really cool! I think it would be useful to have a trash bin folder per user, similar to the way the root folders are currently organized. That way folder's could also be put into the trash bin and would retain their hierarchy. Bookmarks should still have a deleted property to avoid displaying them for searches or similar. And we'd need some kind of auto-cleaning of the trash-bin.

DeathByDenim:
Sounds good. Forked and at work! I'll let you know if I run into any trouble.

DeathByDenim:
Hi @marcelklehr ,

I've been making progress (slowly!). I've been modelling the behaviour of the trash on the Files application. It works now for bookmarks:

https://user-images.githubusercontent.com/5130183/104537475-09325800-55e8-11eb-959d-65b441052a4e.mp4

The trash folder is per user. Deleted bookmarks don't show up anywhere else, including searches, thanks to the new "deleted" property they now have.

Next up are folders. I was going to model that on the Files application again, which does not respect folder structure. Well, in displaying it anyway. When a file is deleted from a folder, it will appear in the root of the trash bin. When restoring, it will go back into the original folder (if it still exist, it not it goes to the root).

For consistency's sake, it would probably be good to go with the same behaviour, but I think it is different from what you described because some of the hierarchy would be lost unless the bookmark is restored. What is your opinion of this?

Thanks!

[...]

PR #1402, opened 2021-02-03
Addition of recycle bin for undeleting bookmarks and folders

DeathByDenim:
This is a work-in-progress pull request. This pull requests currently supports deleting and undeleting bookmarks and root-level folders. Intended to address issue #681.

Things that don't work yet:
- Count of deleted bookmarks if a folder is deleted
- Deleting a subfolder only marks it as deleted, but it remains visible
- Clicking on a folder in deleted bookmark changes trash can icon to home icon
- Shares are completely unaccounted for


marcelklehr:
Awesome!

Here are some thoughts, without having seen the code:

> Count of deleted bookmarks if a folder is deleted

This is tricky. Deleting a folder could mark all descendant bookmarks as deleted as well, but bookmarks can potentially be in multiple folders.

> Deleting a subfolder only marks it as deleted, but it remains visible

This is probably due to the tree entry for said folder. It would be necessary to remove that as well for both bookmarks and folders to not litter the tree with deleted items. In order to still store the old parent somewhere we should probably introduce a trashbin table `trashbin(type, id, parent_folder)`

> Clicking on a folder in deleted bookmark changes trash can icon to home icon

I see. I'm not sure if we actually need that behavior. At least for now it would probably enough to just list the deleted folders without being able to enter them. Otherwise people might be confused about whether if the folder their browsing is deleted or not.

> Shares are completely unaccounted for

So, the behavior we want here would probably be: When a folder is deleted, remove all its shares and public links.

DeathByDenim:
> > Count of deleted bookmarks if a folder is deleted
> 
> This is tricky. Deleting a folder could mark all descendant bookmarks as deleted as well, but bookmarks can potentially be in multiple folders.

Indeed, I had noticed that bookmarks could exist in multiple folder, but until you pointed this out, it hadn't occurred to me that this will making counts even more difficult.
> 
> > Deleting a subfolder only marks it as deleted, but it remains visible
> 
> This is probably due to the tree entry for said folder. It would be necessary to remove that as well for both bookmarks and folders to not litter the tree with deleted items. In order to still store the old parent somewhere we should probably introduce a trashbin table `trashbin(type, id, parent_folder)`

Oh, I see. Yes, that would be a nice solution. I'll look into that.
> 
> > Clicking on a folder in deleted bookmark changes trash can icon to home icon
> 
> I see. I'm not sure if we actually need that behavior. At least for now it would probably enough to just list the deleted folders without being able to enter them. Otherwise people might be confused about whether if the folder their browsing is deleted or not.
It would be handy to know for a user what is inside of a folder before deleting it permanently. However, we can make this a separate addition and get this working first.
> 
> > Shares are completely unaccounted for
> 
> So, the behavior we want here would probably be: When a folder is deleted, remove all its shares and public links.
Yes, my thoughts exactly. Restoring a folder might have unexpected results for a user as the shares wouldn't be restored though. Then again, it may also be unexpected if unintended sharing occurs since there is no explicit sharing action performed by the user.

Thanks! This is helpful.

appinfo/routes.php
```diff
@@ -42,9 +43,12 @@
 		['name' => 'internal_bookmark#import_bookmark', 'url' => '/bookmark/import', 'verb' => 'POST'],
 		['name' => 'internal_bookmark#count_unavailable', 'url' => '/bookmark/unavailable', 'verb' => 'GET'],
 		['name' => 'internal_bookmark#count_archived', 'url' => '/bookmark/archived', 'verb' => 'GET'],
+		['name' => 'internal_bookmark#count_deleted', 'url' => '/bookmark/deleted', 'verb' => 'GET'],
 		['name' => 'internal_bookmark#edit_bookmark', 'url' => '/bookmark/{id}', 'verb' => 'PUT'],
 		['name' => 'internal_bookmark#get_single_bookmark', 'url' => '/bookmark/{id}', 'verb' => 'GET'],
 		['name' => 'internal_bookmark#delete_bookmark', 'url' => '/bookmark/{id}', 'verb' => 'DELETE'],
+		['name' => 'internal_bookmark#delete_bookmark_permanently', 'url' => '/bookmark/permanent/{id}', 'verb' => 'DELETE'],
```
marcelklehr, line 50: can we change the URI for this to `DELETE /bookmark/{id}?permanent=true`

appinfo/routes.php
```diff
@@ -42,9 +43,12 @@
...
+		['name' => 'internal_bookmark#restore_bookmark', 'url' => '/bookmark/restore/{id}', 'verb' => 'PUT'],
```
marcelklehr, line 51: ...and this one to `POST /bookmark/{id}/restore`

appinfo/routes.php
```diff
@@ -56,10 +60,13 @@
 		['name' => 'internal_tags#rename_tag', 'url' => '/tag/{old_name}', 'verb' => 'PUT'],
 		['name' => 'internal_tags#delete_tag', 'url' => '/tag/{old_name}', 'verb' => 'DELETE'],
 		['name' => 'internal_folders#get_folders', 'url' => '/folder', 'verb' => 'GET'],
+		['name' => 'internal_folders#get_folders_with_deleted', 'url' => '/folder/withdeleted', 'verb' => 'GET'],
```
marcelklehr, line 63: This should be a Query param as well.

l10n/nl.js
```diff
@@ -44,6 +44,9 @@ OC.L10N.register(
     "Rename" : "Hernoemen",
     "Move" : "Verplaatsen",
     "Delete" : "Verwijderen",
+    "Permanently delete" : "Definitief verwijderen",
+    "Restore" : "Herstellen",
+    "Search" : "Zoeken",
```
marcelklehr, line 49: Translations are handled on transifex.com. These will likely be overriden :/

lib/Controller/FoldersController.php
```diff
@@ -265,6 +265,34 @@ public function addToFolder($folderId, $bookmarkId): JSONResponse {
 	 * @PublicPage
 	 */
 	public function removeFromFolder($folderId, $bookmarkId): JSONResponse {
+		if (!Authorizer::hasPermission(Authorizer::PERM_EDIT, $this->authorizer->getPermissionsForFolder($folderId, $this->request)) &&
+			!Authorizer::hasPermission(Authorizer::PERM_EDIT, $this->authorizer->getPermissionsForFolder($bookmarkId, $this->request))) {
+			return new JSONResponse(['status' => 'error', 'data' => 'Could not find folder'], Http::STATUS_BAD_REQUEST);
+		}
+		try {
+			$this->bookmarks->delete($bookmarkId);
```
marcelklehr, line 273: This is possibly confusing/fragile, as bookmarks that are in multiple folders would be inside and outside the trashbin at the same time.

marcelklehr (changes requested):

appinfo/routes.php
DeathByDenim: I'm not sure if this is possible. I don't see anything about GET parameters like that in https://docs.nextcloud.com/server/latest/developer_manual/basics/routing.html, nor do I see any URI with `?` in the default Nextcloud apps.
Using `?permanent=true` is simply ignored and the method `internal_bookmark#delete_bookmark` is called instead without any way of grabbing the `permanent` parameter that I can see.
Is there a way to do this?

appinfo/routes.php
DeathByDenim: Done

l10n/nl.js
DeathByDenim: Ah oops. Oh well, I undid that.

lib/Controller/FoldersController.php
DeathByDenim: True, it's tricky that these bookmarks can exist in multiple places. I hadn't thought of that when I initially started this.

I guess your suggestion of a new table `trashbin` would be a better fit for this. This would replace the need for the `deleted` flag I have been using so far for bookmarks and folders. Although that wouldn't solve the confusion aspect of bookmarks existing in both the trashbin and outside. Schrödinger's bookmarks, I suppose!

It would probably make the most sense to have them appear in the trashbin since there were deleted from a certain folder, even if the bookmark may still exist in another folder. They were deleted after all. Not sure what to do if two such bookmarks would have been deleted though. It will only show one in the trashbin. How does one restore that? Two could be shown, I suppose but then the original parent folder would have to be indicated somehow. That might work.

appinfo/routes.php
marcelklehr: You can add a $permanent parameter to delete_bookmark and check that. It gets populated with GET parameters

appinfo/routes.php
DeathByDenim: Done

appinfo/routes.php
DeathByDenim: Thanks, that worked!

marcelklehr:
I think, it would probably sanest to do the following:

Only put a bookmark in the trash if there are no references left in any non-trashed folders. When deleting a folder, move it from its parent folder into the trash bin, keeping all descendants intact. Any descendant bookmarks of the trashed folder, should be checked for whether they are still in other folders that are non-trashed, if they are not, mark those bookmarks as deleted. When restoring a folder, check if any descendant bookmarks have to be revived. When deleting a bookmark directly, mark it as deleted as well, but also put it in the trash bin.

@DeathByDenim What do you think? Are there any loopholes in this?

DeathByDenim:
Hi @marcelklehr ,

Yes, that would work. I do see one loophole though.

There could be a situation where a user deletes a bookmark that also exists in another non-trashed folder and they want to restore it. Since a bookmark won't appear in the trash if there are still references to it, it's not possible to restore that bookmark unless the same bookmark is removed from the other folder as well. (Unless I misunderstood you)

Probably the best way to handle this is to put the deleted bookmark in the trash bin immediately, even if there are still references to it in other folders. Then when permanently deleting the entry from the trash bin, only remove that item from `bookmarks_tree` and leave the other entries intact. That way, a user could always restore a bookmark they just deleted.

A side effect of this is that if a user deletes the same bookmark from both folder A as well as B, it would appear once in the trash bin, but restoring it would restore to both folders A and B. That's probably fine though, since if they wanted to restore only in A, they can delete the bookmark from folder B again and then permanently delete it.

marcelklehr:
Ah yes. So:

When deleting a folder, move it from its parent folder into the trash bin, keeping all descendants intact. Any descendant bookmarks of the trashed folder, should be checked for whether they are still in other folders that are non-trashed, if they are not, mark those bookmarks as deleted. When restoring a folder, check if any descendant bookmarks have to be revived. When deleting a bookmark directly, put it in the trash bin along with the folder it was deleted from (possibly requires two different trash bin tables). When no alive references to a bookmark are left, mark the bookmark as deleted.

DeathByDenim:
Ok, that makes sense. Who knew this would be so complicated! :)

radoeka:
What's needed to finish the merger?

DeathByDenim:
I went on a bit of a detour by giving bookmarks a "deleted" flag, but that failed to account for the fact that the same bookmark can exist in multiple folders. So I'm in to process of moving that "deleted" flag to bookmarks_tree instead. That seems to work, but I still need to fix permanently deleting a bookmark and I haven't tackled the shares yet. Getting there though.

After that, there is the testing!

radoeka:
@DeathByDenim :)

PR #1416, opened 2021-02-09
Implement full-text archiving of HTML pages

### LINKED ISSUE #438: Copy page as an artikel
In last time, i prefer wallabag, because it gives me the ability to read my bookmarks offline. Its a perfekt pocket/read it later solution. It would be also an Archievement to read offline in train or somewhere else bookmarked links. Another cool feature is the archive. I can archive readed articles and thats a very nice new possibility of structure. I often bookmark new cooking receipts and then I know which of them I already cooked. 
I also like the raster view with an example picture of the bookmarked site and a little description of it. Little disadvantage for me using wallabag: The apps in nextcloud are better. 

I hope I can post my little suggestions here and hope that it is not an overenhancement for a simple bookmark app. But in my opinion is overview and a little imagination behind the bookmarks the most important thing to find what you want - even if you dont know it (example cooking receipt, there you can easily choose an receipt what you want by choosing based on preview images.) :smile::heart_eyes::grinning:
The table list of bookmark is so unclear. ;)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/54784431-copy-page-as-an-artikel?utm_campaign=plugin&utm_content=tracker%2F216513&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F216513&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #974: Suggestion: make it easier to close Details Menu
Rather than clicking the  small `X` to close, how about just closing the Details Menu by clicking most anywhere above the bookmark title.
Screenshot_20200416_120909

marcelklehr:
This PR adds full-text archiving of links, allowing offline reading and full-text search.

fixes #438
fixes #974

<details>
<summary>Original version, see below for current version</summary>

Peek 2021-02-09 09-08

</details>

- [x] reset the "last archive date" for all bookmarks

/cc @nextcloud/designers @nextcloud/designlearners


tcitworld:
It would be nice for this feature to be toggled through an admin setting, since it will be quite heavier on the database.

Apart from that reading articles into this kind of sidebar tab is definitely not pleasant, it should open a dedicated view using at least half the screen's width instead. Also the links should be rewritten to open in a new tab.

marcelklehr:
> It would be nice for this feature to be toggled through an admin setting, since it will be quite heavier on the database.

Currently, it's hidden behind the same settings flag as the previews

> Apart from that reading articles into this kind of sidebar tab is definitely not pleasant, it should open a dedicated view using at least half the screen's width instead. Also the links should be rewritten to open in a new tab.

I agree that the tab thing is not a nice solution, but I wasn't sure what would work best, especially since the only suitable component available in nextcloud/vue would be a modal, which is also not that nice, IMHO.

stefan-niedermann:
While i am personally not a friend of modals - it might be a solution here. The [Deck](https://github.com/nextcloud/deck/) app uses modals as well...  
The details tab could have a "View content" button or so which opens the modal :man_shrugging: 

grafik

jancborchardt:
Wow nice, this is a great move on functionality @marcelklehr!

Design-wise, I would also say that a modal is already much better than the sidebar. We can make sure it has a certain width too so that the site is not squished.

(Replacing the content with the website view would probably be a bit confusing, so modal seems better than that too.)

marcelklehr:
Thanks for your feedback, folks! Here's the new version with the modal.

Peek 2021-02-10 13-07

stefan-niedermann:
Some more input to your new screencast:

- Provide "closing"-<kbd>X</kbd> in the top right corner
- Make sure it is closable via <kbd>ESC</kbd>

marcelklehr:
> It would be nice for this feature to be toggled through an admin setting, since it will be quite heavier on the database.

@tcitworld Would this be such a big strain on the db, though? This will not archive any images or media, it will just grab the marked up text, which shouldn't be that much. What are you worried about, specifically?

tcitworld:
> Would this be such a big strain on the db, though? This will not archive any images or media, it will just grab the marked up text, which shouldn't be that much

My concern was about an admin who enables the bookmark app may not expect this feature to be available, and the table size to grow x10 suddenly, because data is much heavier than metadata. :)
So off by default with a feature flag seems a reasonable choice.

florianwgnr:
First of all: This feature looks great! Thank you for your work!

> So off by default with a feature flag seems a reasonable choice.

I was just wondering if the "archive a page"-job is done automatically after enabling this feature or if it has to be triggered manually (per bookmark? all at once?). 
The reason why i am asking: i have a lot of bookmarks that i don't need to archive, either because the server can´t reach these pages (local network) or because they are just not interesting (google.com, youtube.com, ...). 

Just to leave the idea here: maybe an option like "save content" per bookmark in their context menu (or sidebar) could address both of these issues...

marcelklehr:
> I was just wondering if the "archive a page"-job is done automatically after enabling this feature or if it has to be triggered manually (per bookmark? all at once?).

The crawl job runs automatically, every half hour on a hundred bookmarks per run. I'd say it's not too problematic if some of those bookmarks are not articles, they will simply have less useful text content archived. It would probably be a good idea to reset the "last archive date" for all bookmarks when this feature is released, though, as bookmarks are only re-crawled every few months.

marcelklehr:
Related to https://github.com/nextcloud/viewer/issues/794

marcelklehr:
Update:

Peek 2021-03-05 13-30

github-account1111:
Why text only? Media is also important.

marcelklehr:
> Why text only? Media is also important.

Media is not that easy. One thing at a time. :)

florianwgnr:
> What do you think?

I think it looks really nice! :) 

But i´m still wondering if there should be an option/a flag to enable/disable the saving of content. I - for example - just want to archive a few of my bookmarks because most of them are local-links (they won´t work and leave an error) or are not relevant for me.

marcelklehr:
> But i´m still wondering if there should be an option/a flag to enable/disable the saving of content. I - for example - just want to archive a few of my bookmarks because most of them are local-links (they won´t work and leave an error) or are not relevant for me.

Unaccessible links will not cause an error. They will simply not be downloaded.

github-account1111:
Does this work like Evernote's web clipper (a "faithful" copy taken from own browser) or Nunux Keeper (server visits page separately). The latter leads to bad formatting (due to lack of content filtering + not being logged in on certain websites).

marcelklehr:
At the moment the server visits the page separately. I'm open to implementing a way for third-party apps to set the archived content, so that a web clipper could be built. The current solution does have a content filter, though, which is based on the algorithm used by Mozilla's Pocket.

marcelklehr:
Peek 2021-04-12 16-13

Bugsbane:
Came across this rather late, but just wanted to say wow... capturing the text for searching later is such a great feature that I never even thought of wanting. Really nice work on this!

df911:
Is there a plan to also archive media such as images? Can this be included in the roadmap?

marcelklehr:
@df911 Direct links to media files are already archived if you enable archiving in the app's user settings. Archiving media *as part of the webpages* they are in, is tougher. You can open a new issue for that, if you want that :)

PR #1432, opened 2021-02-23
Fix RepairStep: Reinsert lost bookmarks instead of deleting

marcelklehr:

### DIFF
--- a/lib/Migration/OrphanedTreeItemsRepairStep.php
+++ b/lib/Migration/OrphanedTreeItemsRepairStep.php
@@ -112,19 +112,30 @@ class OrphanedTreeItemsRepairStep implements IRepairStep {
 		$output->info("Removed $i orphaned bookmark folders");
 
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('b.id')
+		$qb->select('b.id', 'b.user_id')
 			->from('bookmarks', 'b')
 			->leftJoin('b', 'bookmarks_tree', 't', 'b.id = t.id AND t.type = '.$qb->createPositionalParameter('bookmark'))
 			->where($qb->expr()->isNull('t.id'));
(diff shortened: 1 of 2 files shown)


susinths:
Looks good to me from an upgrade from NC19 and bookmarks v3.4.9 to NC20 and bookmarks v4.1.0 :+1:

radoeka:
Would it be possible to complete the merge?  Only 1 check to go (php7.4-sqlite-stable20).

marcelklehr:
The reason it hasn't been merged, yet is pending code reviews :)

(ping @ChristophWurst, @blizzz, @stijink )

radoeka:
When I'm not mistaken (after reading the code), the reinserted bookmark is stored in the root folder.  
Would it be possible to store it in an orphaned bookmark folder?  And or give it the tag orphaned?

Having it stored in a "Orphaned Bookmarks" folder (and/or tagged with orphaned) it will be much easier to find it and reorganize it back into the folder structure.

stijink:
@marcelklehr I would recommend to add a functional test. Is this possible with nextcloud ?

marcelklehr:
@stijink That's a good idea! There's basic CI for repair steps: https://github.com/nextcloud/bookmarks/blob/master/.github/workflows/repair.yml

This imports a basic folder structure into the app, runs the repair steps and compares the API-returned tree structure from before and after repair steps. Do you think we need a specific test to make sure this legacy migration works?

marcelklehr:
@stijink I've added a test :heart:

marcelklehr:
@radoeka we can iterate over this in the future. For now, let's make sure people don't lose their bookmarks anymore.

marcelklehr:
@ChristophWurst @blizzz Could you take a look at this? @stijink is not available at the moment :)

marcelklehr:
#mergetime

PR #1484, opened 2021-03-30
Dashboard: Add 'frequent bookmarks' widget

marcelklehr:


src/dashboard.js
```diff
@@ -6,21 +6,29 @@
...
-			router,
-			render: h => h(Dashboard),
+			store: new Store(deepClone(store)),
+			render: h => h(DashboardRecent),
+		})
+	})
+	OCA.Dashboard.register('bookmarks.frequent', (el) => {
+		global.BookmarksFrequent = new Vue({
+			el,
+			store: new Store(deepClone(store)),
```
marcelklehr, line 30: We want two separate stores, because the two widgets may be on the page at the same time

src/components/DashboardFrequent.vue
```diff
@@ -0,0 +1,44 @@
...
+<script>
+import { DashboardWidget } from '@nextcloud/vue-dashboard'
+import { generateUrl } from '@nextcloud/router'
+import { actions } from '../store'
+export default {
+	name: 'DashboardFrequent',
+	components: { DashboardWidget },
+	computed: {
+		loading() {
+			return Boolean(this.$store.state.loading.bookmarks)
```
skjnldsv, line 24: I have never seen this used :scream: 
No idea if that is good or bad :grin:

src/components/DashboardFrequent.vue
```diff
@@ -0,0 +1,44 @@
...
+		},
+		items() {
+			return this.$store.state.bookmarks.map(bookmark => ({
```
skjnldsv, line 27: You should not get a state like that, getters are here for that purpose :)

src/dashboard.js
```diff
@@ -6,21 +6,29 @@
...
+import { Store } from 'vuex'
+import deepClone from 'clone-deep'
 
 Vue.mixin(AppGlobal)
 Vue.directive('tooltip', Tooltip)
 
 document.addEventListener('DOMContentLoaded', () => {
 	OCA.Dashboard.register('bookmarks.recent', (el) => {
-		global.Bookmarks = new Vue({
+		global.BookmarksRecent = new Vue({
```
skjnldsv, line 21: Where is this `global` var coming from?

src/main.js
```diff
@@ -17,7 +18,7 @@ Vue.directive('tooltip', Tooltip)
 
 const BookmarksApp = (global.Bookmarks = new Vue({
```
skjnldsv, line 19: Same question for `global`

src/components/DashboardFrequent.vue
marcelklehr: Ah, good point!

src/dashboard.js
marcelklehr: I think webpack polyfills `global` with `window`

src/components/DashboardFrequent.vue
marcelklehr: Fixed

src/components/DashboardFrequent.vue
marcelklehr: I dunno, I use it for type casting. What do you use?

PR #1652, opened 2021-09-21
Typo copyng

JSoko:
Changed typo copyng to copying


marcelklehr:
Thanks for the heads up. We actually decided to rename it from "copying" to "add to folders" and this instance was overlooked.

PR #1660, opened 2021-09-29
Drop symfony dependency

kesselb:


marcelklehr:
Huh. I could have sworn those were still needed :thinking: 

Thanks!

marcelklehr (approved):

kesselb:
Thanks ;) Some Symfony components are already shipped with Nextcloud. In addition we could scope php-cs-fixer and psalm like https://github.com/nextcloud/server/pull/28507 to have cleaner dev dependencies.

PR #1771, opened 2022-03-25
CI: Test with nextcloud 24

marcelklehr:


marcelklehr:
It seems it does not https://github.com/nextcloud/bookmarks/runs/5820678367

nickvergessen:
So, the reason why it doesn't work is because you are installing an older version first which still claims to support oracle.
Only the update test fails, normal unit tests are fine as there you directly install the new version which does not claim to support oracle.

So this is only a theoretical problem:
1. When you publish the fixed version without Oracle support, people can not install the older one anymore as they will be offered the newer one
2. When https://github.com/nextcloud/server/pull/31825 is merged, people that have have a old version installed can update just fine as the added check is skipped when the column already existed
3. When https://github.com/nextcloud/server/pull/31825 is merged new installs will work fine as the check is skipped when oracle is not supported

So to fix it backport https://github.com/nextcloud/bookmarks/commit/e70ccbe3b16b8c9358b6adf31bdebcc5ed874e75 to stable stable3 and stable4 and make a release, then your CI job will pass, but as per above no user should be affected

nickvergessen:
Btw I saw you are having lots of actions. You are currently running every test twice, once for pushing the commit to the branch and once for the pull request.
I recommend limiting pushes to branches master and stable, like we do in all other apps:
https://github.com/nextcloud/notifications/blob/73e4292311c1f7a2e1590eeb86f78023454e5287/.github/workflows/phpunit.yml#L12-L15

PR #1873, opened 2022-08-18
Hide deprecated projects in sidebar by default

Pytal:
Deprecate projects which is superseded by Related resources https://github.com/nextcloud/server/issues/28320


marcelklehr:
:cry:

PR #1877, opened 2022-08-23
Avoid crawling bookmark files responding HTTP error

fengkaijia:
The crawler currently only skipping bookmark links that responded with HTTP 404 status during the cron job. An URL returning a 500 status and a `text/plain` content type will result in an empty txt file being created in the bookmarks folder, such as:

```
< HTTP/1.1 500 Internal Server Error
< Server: [HIDDEN]
< Date: Tue, 23 Aug 2022 02:05:27 GMT
< Content-Length: 0
< Content-Type: text/plain; charset=utf-8
```

This PR marks a bookmark URL with any 4xx and 5xx errors as unavailable to stop this behavior.

### DIFF
--- a/lib/Service/CrawlService.php
+++ b/lib/Service/CrawlService.php
@@ -95,7 +95,7 @@ class CrawlService {
 				'read_timeout' => self::READ_TIMEOUT,
 				'http_errors' => false
 			]);
-			$available = $resp ? $resp->getStatusCode() !== 404 : false;
+			$available = $resp ? $resp->getStatusCode() < 400 : false;
 		} catch (Exception $e) {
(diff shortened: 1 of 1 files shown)


marcelklehr:
Mh, the specificty to only mark 404s as unavailable is to avoid false positives, as 500 errors are usually temporary and 401/403 usually require authentication that the user may have but the server doesn't.

marcelklehr:
Could you change this line instead? https://github.com/nextcloud/bookmarks/pull/1877/files#diff-f005da827c197199c57d877a41587d363804338aebb685852263f0bfcecb7c68R104

PR #1889, opened 2022-09-03
fix service worker

### LINKED ISSUE #1078: Implement installing as webAPK?
> With v3.2.0 released you can now install the nextcloud bookmarks web app as a progressive web app on your android or iOS phone.

Thank you very much for implementing this feature. It is working great! 

Would it be possible to feature the installment as "WebAPK"?
This would allow the PWA to become much more independent from the Browser and some other advantages.

Here are the docs:
https://developers.google.com/web/fundamentals/integration/webapks

> Being installed via an APK makes it possible for your app to show up in the app launcher, in Android's app settings and to register a set of intent filters.

_Originally posted by @szaimen in https://github.com/nextcloud/bookmarks/issues/1051#issuecomment-643833023_

szaimen:
Found via https://github.com/search?q=org%3Anextcloud+noRewrite&type=code

Close https://github.com/nextcloud/bookmarks/issues/1078

Disclaimer: I wasn't able to test this but I think it should work.

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -34,10 +34,9 @@ const BookmarksApp = (global.Bookmarks = new Vue({
 }))
 
 if ('serviceWorker' in navigator) {
-	navigator.serviceWorker.register(
-		generateUrl('/apps/bookmarks/service-worker', {}, {
-			noRewrite: true,
-		}) + '.js')
+	navigator.serviceWorker.register(generateUrl('/apps/bookmarks/service-worker.js', {}, {
+		noRewrite: true,
+	}))
 		.then(() => {
 			console.info('ServiceWorker registered')
 		})
--- a/src/main.js
+++ b/src/main.js
@@ -34,9 +34,10 @@ const BookmarksApp = (global.Bookmarks = new Vue({
 }))
 
 if ('serviceWorker' in navigator) {
-	navigator.serviceWorker.register(generateUrl('/apps/bookmarks/service-worker.js', {}, {
-		noRewrite: true,
-	}))
+	navigator.serviceWorker.register(
+		generateUrl('/apps/bookmarks/service-worker', {}, {
+			noRewrite: true,
+		}) + '.js')
 		.then(() => {
 			console.info('ServiceWorker registered')
 		})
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -34,7 +34,11 @@ const BookmarksApp = (global.Bookmarks = new Vue({
 }))
 
 if ('serviceWorker' in navigator) {
-	navigator.serviceWorker.register(generateUrl('/apps/bookmarks/service-worker.js'))
+	const webRootWithIndexPHP = getRootUrl() + '/index.php'
+	const doesURLContainIndexPHP = window.location.pathname.startsWith(webRootWithIndexPHP)
+	navigator.serviceWorker.register(generateUrl('/apps/bookmarks/service-worker.js', {}, {
+		noRewrite: doesURLContainIndexPHP,
```
szaimen, line 40:
```suggestion
		noRewrite: doesURLContainIndexPHP,
```
we may need to invert the logic here to make it work. not sure though...

szaimen: or we simply always set it to true, WDYT?

szaimen: Will do this...

szaimen: done

szaimen:
ping @marcelklehr

szaimen:
@marcelklehr I think this should work now. However not really sure how to test since my test instance uses a self-signed certificate which makes service-workers unsupported by definition.

src/main.js
```diff
@@ -34,7 +34,11 @@ const BookmarksApp = (global.Bookmarks = new Vue({
 }))
 
 if ('serviceWorker' in navigator) {
-	navigator.serviceWorker.register(generateUrl('/apps/bookmarks/service-worker.js'))
+	navigator.serviceWorker.register(generateUrl('/apps/bookmarks/service-worker.js', {}, {
+		noRewrite: true,
+	}), {
+		scope: generateUrl('/apps/bookmarks'),
```
szaimen, line 40: However assets might be loaded through a different route so not sure if this would really work. Thoughts?

szaimen: based on @artonge does this indeed make sense: https://github.com/nextcloud/photos/pull/1294#discussion_r980002298

marcelklehr (approved):

marcelklehr:
Tested and can confirm it works! Thanks :)

Issue #1890, opened 2022-09-06
Backup frequency setting


JackRoublard:
**Is your feature request related to a problem? Please describe.**
Thank you very much for implementing automatic backups ! Although this is undeniably a must-have and very welcome feature, I remain a bit frustrated by the rigidity of the chosen implementation :

> I [...] have implemented a regular backup, keeping backups of
> 
>     * every day for the past 7 days
> 
>     * every Monday for the past 5 weeks
> 
>     * every first Monday for the past 6 months

**Describe the solution you'd like**
Would it be possible to add an option to set the backup frequency ? Maybe this will have to wait for #1798 to be solved, in which case I suggest the following.

**Describe alternatives you've considered**
As a simpler alternative, letting the user set a cap for the number of generated backup files would fit a lot of use cases, I believe. 

**Additional context**
Related to #1798, whose outcome might influence this issue.

PR #1973, opened 2023-02-16
Optimize backup job

### LINKED ISSUE #1798: Use versioning for backups instead of custom hard-coded schedule
#1318 nextcloud/bookmarks@0d7ce3d introduced a backup feature, which got enabled by default *(sigh; not going over issues with that now. Should be obvious by now as it got mentioned in other issues as well)*.
#1779 & #1781 asked for a feature to disable these backups. _(which is now possible for a single user)_

> this needs to be a global admin option to disable it for everyone.

https://github.com/nextcloud/bookmarks/issues/1779#issuecomment-1100946407 wasn't implemented [yet, but the issue was closed already]

So, I'm opening this issue for a global admin option^. And…

> I've revised this decision now, and have implemented a regular backup, keeping backups of
> 
> - every day for the past 7 days
> - every Monday for the past 5 weeks
> - every first Monday for the past 6 months

https://github.com/nextcloud/bookmarks/issues/1318#issuecomment-1065080553 
It's nice to have some sort of backup strategy, but why aren't you using existing versioning in nextcloud (which can be configured by the admin). [Quick-Docs reference](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/file_versioning.html)

Hence I propose updating the same file, instead of creating&deleting yourself. This would allow much better user-controlled backup behavior.

edit.: fixed issue linking;

### LINKED ISSUE #1972: BackupJob runs for too long!
**Describe the bug**
The BackupJob runs for 45 minutes just to create a backup for a batch of ten users. This is very problematic as it is set to run once every five minutes by default so it blocks other  jobs from executing properly.

**Expected behavior**
- Backup Job should be quick
- Batch size should be set by admins
- Job frequency should be set by admins

akhil1508:
Closes https://github.com/nextcloud/bookmarks/issues/1972 and https://github.com/nextcloud/bookmarks/issues/1890

Partially resolves https://github.com/nextcloud/bookmarks/issues/1798 by allowing to disable the backup job

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ npm run build
 
 ## Configuration
 
-### Backup Job Frequency and batch size
+### Backup Job
+- To disable the bookmarks backup job(it is enabled by default), set `bookmarks.backupjob.enabled` to a boolean value of `false` in your `config.php`
 - To set the bookmarks backup job frequency, set `bookmarks.backupjob.interval` in your `config.php` to an integer value in seconds
   - The default frequency is 5 minutes i.e. 300 seconds
 - To set the bookmarks backup batch size, set `bookmarks.backupjob.batch_size` in your `config.php` to an integer value
--- a/README.md
+++ b/README.md
@@ -86,6 +86,15 @@ npm install
 npm run build
 ```
 
+## Configuration
+
+### Backup Job Frequency and batch size
+- To set the bookmarks backup job frequency, set `bookmarks.backupjob.interval` in your `config.php` to an integer value in seconds
+  - The default frequency is 5 minutes i.e. 300 seconds
+- To set the bookmarks backup batch size, set `bookmarks.backupjob.batch_size` in your `config.php` to an integer value
+  - The default batch size is 10 
+
+
 ## API
 
 This app exposes a public REST API that third-party clients can interface with.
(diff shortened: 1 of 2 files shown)


akhil1508:
@marcelklehr FYI

Also thanks for your work on this app, it is a life-saver :blush:

marcelklehr:
Hi @akhil1508 
thank you for stepping up to fix this. I was thinking instead of adding setings to fine-tune the backup job, can we simply add a constraint so it doesn't run longer than 5minutes?

akhil1508:
>  I was thinking instead of adding setings to fine-tune the backup job, can we simply add a constraint so it doesn't run longer than 5minutes?

@marcelklehr In our use case, we'd definitely like to run it once a day and for all users instead of this kind of batching.. which is why I feel this is a good idea.

marcelklehr:
> instead of this kind of batching

As I see it there is no way around this kind of batching because Nextcloud will run cron jobs every 5 minutes. If a cron job takes much longer than that it can cause hiccups, because it blocks other jobs from running. The point of this 5min interval is definitely not to *do* backups every 5minutes, but to be *able* to do them for as long as necessary.

akhil1508:
@marcelklehr Closing this as you applied changes already :+1:

PR #1981, opened 2023-02-22
BackupJob must run for a maximum of MAX_RUN_TIME

akhil1508:
- Sign was wrong in master
- Better to separate interval from job run time constant imo


akhil1508:
@marcelklehr FYI

marcelklehr:
Thank you :hearts:

PR #2037, opened 2023-07-07
show context menu for shared folders without edit permission

### LINKED ISSUE #2036: Context menu for shared folders only visible with edit permission
**Describe the bug**
When I share a folder, it is placed in the root of the other user. They don't see the context menu (three dots) and can therefore not move the folder somewhere else. Even if they have the permission to re-share it, they can not do that. Only if they have the edit permission, the context menu shows up.

**To Reproduce**
1. Share a folder without additional permissions (read-only)
2. Log in as the other user
3. Context menu is not visible

**Expected behavior**
The other user should see the context menu and be able to (locally) rename and move the folder. Maybe even delete (i.e. unshare) it.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 114

**Server (please complete the following information):**
 - OS: Ubuntu 22
 - HTTP server: Apache
 - Database: MySQL
 - PHP version: 7.2
 - Nextcloud version: 25.0.7
 - Bookmarks app version: 12.1.0
 - Nextcloud external user backend: none

Plunts:
Closes #2036

### DIFF
--- a/lib/Controller/FoldersController.php
+++ b/lib/Controller/FoldersController.php
@@ -276,7 +276,8 @@ class FoldersController extends ApiController {
 	 * @PublicPage
 	 */
 	public function deleteFolder($folderId): JSONResponse {
-		if (!Authorizer::hasPermission(Authorizer::PERM_EDIT, $this->authorizer->getPermissionsForFolder($folderId, $this->request))) {
+		$perms = $this->authorizer->getPermissionsForFolder($folderId, $this->request);
+		if (!Authorizer::hasPermission(Authorizer::PERM_READ, $perms)) {
 			return new JSONResponse(['status' => 'error', 'data' => 'Unauthorized'], Http::STATUS_FORBIDDEN);
 		}
 
@@ -285,7 +286,11 @@ class FoldersController extends ApiController {
 			return new JSONResponse(['status' => 'success']);
 		}
 		try {
-			$this->folders->deleteSharedFolderOrFolder($this->authorizer->getUserId(), $folderId);
+			if (Authorizer::hasPermission(Authorizer::PERM_EDIT, $perms)) {
+				$this->folders->deleteSharedFolderOrFolder($this->authorizer->getUserId(), $folderId);
+			} else {
+				$this->folders->deleteSharedFolder($this->authorizer->getUserId(), $folderId);
+			}
 			return new JSONResponse(['status' => 'success']);
 		} catch (UnsupportedOperation $e) {
 			return new JSONResponse(['status' => 'error', 'data' => 'Unsupported operation'], Http::STATUS_INTERNAL_SERVER_ERROR);
@@ -308,14 +313,19 @@ class FoldersController extends ApiController {
 	 * @PupblicPage
 	 */
 	public function editFolder($folderId, $title = null, $parent_folder = null): JSONResponse {
-		if (!Authorizer::hasPermission(Authorizer::PERM_EDIT, $this->authorizer->getPermissionsForFolder($folderId, $this->request))) {
+		$perms = $this->authorizer->getPermissionsForFolder($folderId, $this->request);
+		if (!Authorizer::hasPermission(Authorizer::PERM_READ, $perms)) {
 			return new JSONResponse(['status' => 'error', 'data' => 'Unauthorized'], Http::STATUS_FORBIDDEN);
 		}
 		if ($parent_folder !== null) {
 			$parent_folder = $this->toInternalFolderId($parent_folder);
 		}
 		try {
-			$folder = $this->folders->updateSharedFolderOrFolder($this->authorizer->getUserId(), $folderId, $title, $parent_folder);
+			if (Authorizer::hasPermission(Authorizer::PERM_EDIT, $perms)) {
+				$folder = $this->folders->updateSharedFolderOrFolder($this->authorizer->getUserId(), $folderId, $title, $parent_folder);
+			} else {
+				$folder = $this->folders->updateSharedFolder($this->authorizer->getUserId(), $folderId, $title, $parent_folder);
+			}
 			return new JSONResponse(['status' => 'success', 'item' => $this->_returnFolderAsArray($folder)]);
 		} catch (DoesNotExistException $e) {
 			return new JSONResponse(['status' => 'error', 'data' => 'Could not find folder'], Http::STATUS_BAD_REQUEST);
--- a/lib/Service/FolderService.php
+++ b/lib/Service/FolderService.php
@@ -183,6 +183,21 @@ class FolderService {
 	 * @throws MultipleObjectsReturnedException
 	 * @throws UnsupportedOperation
 	 */
+	public function deleteSharedFolder($userId, $folderId): void {
+		/**
+		 * @var $sharedFolder SharedFolder
+		 */
+		$sharedFolder = $this->sharedFolderMapper->findByFolderAndUser($folderId, $userId);
+		$this->treeMapper->deleteEntry(TreeMapper::TYPE_SHARE, $sharedFolder->getId());
+	}
+
+	/**
+	 * @param $userId
+	 * @param $folderId
+	 * @throws DoesNotExistException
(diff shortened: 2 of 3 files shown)


marcelklehr:
Hey!

Really cool that you stepped up and took a shot at this!

lib/Service/FolderService.php
```diff
@@ -226,19 +230,21 @@ public function deleteShare($shareId): void {
...
 	 */
-	public function updateSharedFolderOrFolder($userId, $folderId, $title = null, $parent_folder = null) {
+	public function updateSharedFolderOrFolder($userId, $folderId, $title = null, $parent_folder = null, $perms = Authorizer::PERM_NONE) {
 		/**
 		 * @var $folder Folder
 		 */
 		$folder = $this->folderMapper->find($folderId);
 
-		if ($userId !== null || $userId !== $folder->getUserId()) {
+		if ($userId === null || $userId !== $folder->getUserId()) {
```
marcelklehr, line 247: Any reason for this change? I'm not convinced by the original line but neither by this new version. $userId shouldn't be null here, I believe because we need it to find the sharedFolder. Maybe my suggestion makes sense?

```suggestion
		if ($userId !== null && $userId !== $folder->getUserId()) {
```

Plunts: I just came across this line and thought it is suspicious. I guess it works, because there will be no shares for the owner and so the code doesn't do anything in that case. Your version makes sense, if there is no way for unregistered users to edit folders.

lib/Controller/FoldersController.php
```diff
@@ -308,14 +308,12 @@ public function deleteFolder($folderId): JSONResponse {
 	 * @PupblicPage
 	 */
 	public function editFolder($folderId, $title = null, $parent_folder = null): JSONResponse {
-		if (!Authorizer::hasPermission(Authorizer::PERM_EDIT, $this->authorizer->getPermissionsForFolder($folderId, $this->request))) {
-			return new JSONResponse(['status' => 'error', 'data' => 'Unauthorized'], Http::STATUS_FORBIDDEN);
```
marcelklehr, line 312: Curious: What lead you to the conclusion that you needed to refactor these lines? Currently the PR fails a security test. I'd rather we keep authorization in the controller.

Plunts: The service has a "fall-through" (with the no-op catch) that seems to be used to allow renaming the actual folder if it's a subfolder of a share. This is behaivour is of course undesired when it's read-only. I explicitly tested that by allowing the action in the UI no matter the permissions and the backend accepted me to change the name of a subfolder. The alternative was to create a new method that only renames shares, but I didn't like the code dublication that comes with it. 

Maybe we could check for READ-access here to have at least some check and then leave the additional distinction in the service as it is now?

marcelklehr: I believe we should change the following line in the Authorizer instead of pushing authorization down to the service layer: https://github.com/nextcloud/bookmarks/blob/master/lib/Service/Authorizer.php#L291

Plunts: What exactly do you want to change there? If we remove the `hasDescendant` check it would completly disable permission inheritance, wouldn't it? So if someone shares *with* write permissions that would only be true for the folder itself, not subfolders. That isn't the behaivour I'd expect.

New idea: I think with a little bit of refactoring, we could split the `updateSharedFolderOrFolder` method into an addtional `updateSharedFolder`. Then we could call that from the controller if there are only READ permissions. Would that pass the security test?

```php
if (Authorizer::hasPermission(Authorizer::PERM_EDIT, $perms)) {
  $folder = $this->folders->updateSharedFolderOrFolder($this->authorizer->getUserId(), $folderId, $title, $parent_folder);
} else if (Authorizer::hasPermission(Authorizer::PERM_READ, $perms)) {
  $folder = $this->folders->updateSharedFolder($this->authorizer->getUserId(), $folderId, $title, $parent_folder);
}
```

Plunts: I changed the check like I suggested earlier, please have a look if that suits the security requirements.

marcelklehr: > If we remove the hasDescendant check it would completly disable permission inheritance, wouldn't it? 

I was focusing on the `$share->getFolderId() === $folderId` part. This is the part that identifies a share as the source of a folder's permissions. If we changed this to

```php
foreach ($shares as $share) {
	if ($this->treeMapper->hasDescendant($share->getFolderId(), TreeMapper::TYPE_FOLDER, $folderId)) {
		return $this->getMaskFromFlags($share->getCanWrite(), $share->getCanShare());
	}
	if ($share->getFolderId() === $folderId) {
		$sharedFolder = $this->sharedFolderMapper->findByShareAndUser($share->getId(), $userId);
		$parentFolder = $this->treeMapper->findParentOf(TreeMapper::TYPE_SHARE, $sharedFolder->getId());
		return $this->getUserPermissionsForFolder($userId, $parentFolder->getId())
	}
}
```

lib/Controller/FoldersController.php
```diff
@@ -285,7 +286,11 @@ public function deleteFolder($folderId): JSONResponse {
 			return new JSONResponse(['status' => 'success']);
 		}
 		try {
-			$this->folders->deleteSharedFolderOrFolder($this->authorizer->getUserId(), $folderId);
+			if (Authorizer::hasPermission(Authorizer::PERM_EDIT, $perms)) {
+				$this->folders->deleteSharedFolderOrFolder($this->authorizer->getUserId(), $folderId);
+			} else {
+				$this->folders->deleteSharedFolder($this->authorizer->getUserId(), $folderId);
```
marcelklehr, line 292: userId may not always be set

marcelklehr: If it's not set I guess we should just abort with Unauthorized

lib/Controller/FoldersController.php
marcelklehr: mmmh, forget the code example above... I'm trying to get to the bottom of this now.

lib/Controller/FoldersController.php
Plunts: It was used without checks in the version before, so I just left it as it was. But it may be the reason for the null-check in the FolderService that you commented on first 😉

lib/Controller/FoldersController.php
Plunts: So your idea is to let `getUserPermissionsForFolder` return `PERM_WRITE` when the share is accessed directly? Sounds a bit like it would allow the user to also add bookmarks in the shared folder, which would be unintended. Also I would have a bad feeling to call the kind of powerful `updateSharedFolderOrFolder` and risk an unintended fall-through. The change I pushed is much less invasive and nearly riskless.

marcelklehr: I agree that your change is less risky, but I think this is ultimately a permissions issue. Basically, we need to distinguish between RENAME, MOVE and EDIT CONTENTS permissions. Otherwise things get hairy when we start to do different things based on where a permission was used.

Plunts: I agree, that would be the best variant. But it's way above what I feel entitled to do in my first PR 😄 But one more counter-argument: I feel like the user should **always** be allowed to move/rename a share. That's not something the owner is deciding, because it's none of their business. So it would be more like an differentiation in the Authorizer, nothing that is configured via the API and technically you also don't even have to store it in the DB.

marcelklehr:
I've since come up with this PR to tackle your issue: https://github.com/nextcloud/bookmarks/pull/2047

lib/Controller/FoldersController.php
marcelklehr: Completely agree. Thank you nonetheless for your Pull request which caused me to think more deeply about this problem and helped me come up with a good solution :blue_heart:

Plunts: Glad that it helped anyways 🙂 I guess you can close this PR then. Since I think I found one more bug, you'll hear of me again if I tracked it down 😉

PR #2199, opened 2024-07-22
feat(previews): use secure version of screeenly

stefanzweifel:
Hi there 👋 

Maintainer/host of screeenly.com here. This PR changes the default API URL used by the `ScreeenlyBookmarkPreviewer` to use the version of screeenly that is secured with a SSL certificate.

This version has been available for years, but has never really been documented. To prevent breaking changes, the default domain (screeenly.com) has never been issued a SSL certificate.
The app on secure.screeenly.com is exactly the same as on screeenly.com. User with existing tokens can switch to this version without rotating their API tokens.

The non-secure version of screeenly is being deprecated in the next few months. I will inform users of the app directly once I've decided on an exact date.


marcelklehr:
Thank you for reaching out! :)

marcelklehr (approved):

PR #2271, opened 2025-02-21
feat: Add a ContextChat provider

marcelklehr:


marcelklehr:
Mmh, I'm getting hundreds of 

```
"PUT /loadSources HTTP/1.1" 200 OK
```

but context_chat:stats still shows:

```
Indexed documents: array (
  'bookmarks__bookmarks' => 1,
```

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,132 @@
...
+
+	public function register(): void {
+		$this->contentManager->registerContentProvider($this->getAppId(), $this->getId(), self::class);
+	}
+
+	/**
+	 * The ID of the provider
+	 *
+	 * @return string
+	 * @since 1.1.0
```
kyteinsky, line 79: 15.x.x here and other places

lib/ContextChat/ContextChatProvider.php
```diff
@@ -0,0 +1,132 @@
...
+					$bookmark->getTitle(),
+					$bookmark->getTextContent(),
+					'Website',
+					new \DateTime('@' . $bookmark->getLastmodified()),
+					[$user->getUID()]
+				);
+				if (count($items) < 25) {
+					continue;
+				}
+				$this->contentManager->submitContent($this->getAppId(), $items);
```
kyteinsky, line 128: why the 25 item limit?

also, the submitContent should be outside the loop.

kyteinsky (commented):
works nicely for addition and deletions of bookmarks!

lib/ContextChat/ContextChatProvider.php
kyteinsky: ah sorry I misread it. The items array should still be cleared up when submitContent is called and checked once outside the loop.

kyteinsky:
> Mmh, I'm getting hundreds of

submit content has a separate bg job that has a batch size of 20 so might be due to sheer no. of bookmarks?
what does ccb.log say?

marcelklehr:
Yeah, it works now, thanks for having a look :) I think I'll add a user setting for this, though, since people are skeptical of AI these days...

Issue #2361, opened 2026-02-18
SQL Error "SQLSTATE[23505]: Unique violation: 7 ERROR" when updating v6.0.1 -> v6.1.0


v3DJG6GL:
**Describe the bug**

**To Reproduce**
I get this error when updating from v6.0.1 -> v6.1.0:
```
occ upgrade
Setting log level to debug
Turned on maintenance mode
Updating database schema
Updated database
Updating <bookmarks> ...
Exception: Database error when running migration 016002000Date20260201124723 for app bookmarks
An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "bookmarks_uniq_url"
DETAIL:  Key (user_id, url_hash)=(<REDACTED>, d7b8fc8c7ad5a9a55d81525c209311a6) already exists.
Update failed
Maintenance mode is kept active
Resetting log level
```

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS X]
 - Browser: [e.g. Chrome, Safari]
 - Version: [e.g. 22]

**Server (please complete the following information):**
- OS: TrueNAS SCALE 25.10.1
- CPU: 2x Intel Xeon Silver 4316 (2.30GHz|20C|40T)
- RAM: 377.5 GiB
- GPU: NVIDIA RTX 4000 SFF Ada Gen (20 GiB)
- Network (LAN): 10Gbit/s
- Storage (SSD, RAIDZ2): 4x Samsung PM9A3 (3840 GB|U.2|NVMe|Read: 6900 MB/s|Write: 4100 MB/s|IOPS 4K: Read: 1000K Write: 180K)
- Storage (HDD, RAIDZ2): 8x Seagate Exos X20 (20 TB|3.5"|SATA 6Gb/s|7200 RPM|Transfer Rate: 285 MB/s|IOPS 4K: Read: 168 Write: 550)
 - HTTP server: N/A
 - Database: postgres:16-alpine
 - PHP version: v8.4.16
 - Nextcloud version: v32.0.6 (Linuxserver.io Docker) 
 - Bookmarks app version: v6.0.1 -> v6.1.0
 - Activated Nextcloud Apps: N/A
 - Nextcloud external user backend: N/A

**Nextcloud configuration:**
```
occ config:list system
Nextcloud or one of the apps require upgrade - only a limited number of commands are available
[...]

PR #2362, opened 2026-02-18
fix(Migration): Merge per-user-database-duplicate bookmarks before hashing URLs

marcelklehr:
doesn't touch duplicates count in the UI

see #2361


marcelklehr:
Weirdly, Floccus integration tests are not failing on master but failing on this PR, although it only touches the migrations :thinking:

PR #2416, opened 2026-04-30
CrawlService: Update UA_FIREFOX const to recent version

iasdeoupxe:
Hasn't been updated since back then in 2021 with #1518 so just bumping to a more recent version. Could also help as some sites may even block/reject such outdated versions.

Also dropped Ubuntu, Firefox is not sending this anymore since quite some time and probably not really required as well.


marcelklehr (approved):
Ah, nice! Thank you!

PR #2424, opened 2026-05-09
fix: Update manifest link generation to use IURLGenerator

prplecake:
Fixes app for NC34+

### DIFF
--- a/lib/AugmentedTemplateResponse.php
+++ b/lib/AugmentedTemplateResponse.php
@@ -20,7 +20,7 @@ use OCP\IURLGenerator;
 class AugmentedTemplateResponse extends TemplateResponse {
 	public function render() {
 		$return = parent::render();
-		preg_replace('/<link rel="manifest" href="(.*?)">/i', '<link rel="manifest" href="' . OC::$server->get(IURLGenerator::class)->linkToRouteAbsolute('bookmarks.web_view.manifest') . '">', $return);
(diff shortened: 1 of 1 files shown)


lib/AugmentedTemplateResponse.php
```diff
@@ -19,7 +20,7 @@
 class AugmentedTemplateResponse extends TemplateResponse {
 	public function render() {
 		$return = parent::render();
-		preg_replace('/<link rel="manifest" href="(.*?)">/i', '<link rel="manifest" href="' . OC::$server->getURLGenerator()->linkToRouteAbsolute('bookmarks.web_view.manifest') . '">', $return);
+		preg_replace('/<link rel="manifest" href="(.*?)">/i', '<link rel="manifest" href="' . OC::$server->get(IURLGenerator::class)->linkToRouteAbsolute('bookmarks.web_view.manifest') . '">', $return);
```
marcelklehr:
```suggestion
		preg_replace('/<link rel="manifest" href="(.*?)">/i', '<link rel="manifest" href="' . \OCP\Server::get(IURLGenerator::class)->linkToRouteAbsolute('bookmarks.web_view.manifest') . '">', $return);
```

Let's use the official OCP interface as well and it's good to go :) Thank you for the contribution!

marcelklehr:
I've incorporated this change into my PR for 34 compatibility. There were some more issues it turns out :) Thank you nonetheless!

